PR #2, opened 2021-01-29
implement file_handler interface

jake-chandler:
There are a still a few minor changes that might be needed, but I believe it's functionally sound

### DIFF
--- a/project1/server/file_handler/CMakeLists.txt
+++ b/project1/server/file_handler/CMakeLists.txt
@@ -1,2 +1,2 @@
-add_library(file_handler file_handler.h my_file_handler.h my_file_handler.cpp)
+add_library(file_handler my_file_handler.cpp)
 set_target_properties(file_handler PROPERTIES LINKER_LANGUAGE CXX)
--- a/project1/server/file_handler/CMakeLists.txt
+++ b/project1/server/file_handler/CMakeLists.txt
@@ -1,2 +1,2 @@
-add_library(file_handler file_handler.h)
+add_library(file_handler file_handler.h my_file_handler.h my_file_handler.cpp)
 set_target_properties(file_handler PROPERTIES LINKER_LANGUAGE CXX)
--- a/project1/server/file_handler/my_file_handler.cpp
+++ b/project1/server/file_handler/my_file_handler.cpp
@@ -1,7 +1,3 @@
-//
-// Created by jake on 1/25/21.
-//
-
 #include "my_file_handler.h"
 #include <bits/stdc++.h>
 
--- a/project1/server/file_handler/my_file_handler.cpp
+++ b/project1/server/file_handler/my_file_handler.cpp
@@ -4,13 +4,8 @@
 namespace server::file_handler{
   bool MyFileHandler::Delete(const std::string &filename) {
 
-    //convert filename to char*
-    int n = filename.length();
-    char fileName[n+1];
-    strcpy(fileName, filename.c_str());
-
-    //delete file
-    return remove(fileName);
+    return remove(filename.c_str());
+    
   }//Delete
   bool MyFileHandler::Put(const std::string &filename, const std::vector<uint8_t> &contents) {
 
@@ -18,41 +13,26 @@ namespace server::file_handler{
     std::string str_contents(contents.begin(), contents.end());
 
     //open file
-    try {this->os.open(filename);}
-    catch (std::exception &e) {
-      this->os.close();
-      return false;
-    }
+    this->os.open(filename);
 
     //write contents to file
-    try{this->os << str_contents;}
-    catch (std::exception &e){
-      this->os.close();
-      return false;
-    }
+    this->os << str_contents;
 
     //close file
     this->os.close();
+    
     return true;
   }//Put
   bool MyFileHandler::MakeDir(const std::string &name) {
-    //convert directory name to char*
-    int n = name.length();
-    char dir[n+1];
-    strcpy(dir, name.c_str());
-
-    if (mkdir(dir,0777)==-1) return false;
-    else return true;
+    
+    return mkdir(name.c_str(),0777)==-1;
+    
   }//MakeDir
+  
   bool MyFileHandler::ChangeDir(const std::string &sub_folder) {
-
-    //convert subdirectory name to char*
-    int n = sub_folder.length();
-    char dir[n+1];
-    strcpy(dir, sub_folder.c_str());
-
-    if(chdir(dir)==-1) return false;
-    else return true;
+    
+    return chdir(sub_folder.c_str());
+    
   }//ChangeDir
   std::vector<uint8_t> MyFileHandler::Get(const std::string &filename) const {
     std::ifstream stream;
@@ -61,7 +41,7 @@ namespace server::file_handler{
 
     //get contents of file, convert to byte vector
     std::string str_contents( (std::istreambuf_iterator<char>(stream) ),
-                         (std::istreambuf_iterator<char>()    ) );
+                         (std::istreambuf_iterator<char>()));
 
     std::vector<uint8_t> vec_contents(str_contents.begin(),str_contents.end());
 
@@ -71,15 +51,14 @@ namespace server::file_handler{
   }//Get
 
   bool MyFileHandler::UpDir() {
-    if(chdir("..")==1) return false;
-    else return true;
+    return chdir("..")!=-1;
   }//UpDir
 
   std::vector<std::string> MyFileHandler::List() const {
     std::vector<std::string> list;
     for (const auto & file : std::filesystem::directory_iterator(GetCurrentDir())) {
       std::string str(file.path());
-      str.erase(0,GetCurrentDir().length());
+      str.erase(0,GetCurrentDir().length()+1);
       list.push_back(str);
     }
 
--- a/project1/server/file_handler/my_file_handler.cpp
+++ b/project1/server/file_handler/my_file_handler.cpp
@@ -9,23 +9,24 @@ namespace server::file_handler{
   }//Delete
   bool MyFileHandler::Put(const std::string &filename, const std::vector<uint8_t> &contents) {
 
+    std::ofstream stream;
     //convert contents to string format
     std::string str_contents(contents.begin(), contents.end());
 
     //open file
-    this->os.open(filename);
+    stream.open(filename);
 
     //write contents to file
-    this->os << str_contents;
+    stream << str_contents;
 
     //close file
-    this->os.close();
+    stream.close();
     
     return true;
   }//Put
   bool MyFileHandler::MakeDir(const std::string &name) {
     
-    return mkdir(name.c_str(),0777)==-1;
+    return mkdir(name.c_str(),0777)==0;
     
   }//MakeDir
   
--- a/project1/server/file_handler/my_file_handler.cpp
+++ b/project1/server/file_handler/my_file_handler.cpp
@@ -2,7 +2,7 @@
 #include <bits/stdc++.h>
 
 namespace server::file_handler{
-  bool file_handler::MyFileHandler::Delete(const std::string &filename) {
+  bool MyFileHandler::Delete(const std::string &filename) {
 
     //convert filename to char*
     int n = filename.length();
@@ -12,7 +12,7 @@ namespace server::file_handler{
     //delete file
     return remove(fileName);
   }//Delete
-  bool file_handler::MyFileHandler::Put(const std::string &filename, const std::vector<uint8_t> &contents) {
+  bool MyFileHandler::Put(const std::string &filename, const std::vector<uint8_t> &contents) {
 
     //convert contents to string format
     std::string str_contents(contents.begin(), contents.end());
@@ -35,7 +35,7 @@ namespace server::file_handler{
     this->os.close();
     return true;
   }//Put
-  bool file_handler::MyFileHandler::MakeDir(const std::string &name) {
+  bool MyFileHandler::MakeDir(const std::string &name) {
     //convert directory name to char*
     int n = name.length();
     char dir[n+1];
@@ -44,7 +44,7 @@ namespace server::file_handler{
     if (mkdir(dir,0777)==-1) return false;
     else return true;
   }//MakeDir
-  bool file_handler::MyFileHandler::ChangeDir(const std::string &sub_folder) {
+  bool MyFileHandler::ChangeDir(const std::string &sub_folder) {
 
     //convert subdirectory name to char*
     int n = sub_folder.length();
@@ -54,7 +54,7 @@ namespace server::file_handler{
     if(chdir(dir)==-1) return false;
     else return true;
   }//ChangeDir
-  std::vector<uint8_t> file_handler::MyFileHandler::Get(const std::string &filename) const {
+  std::vector<uint8_t> MyFileHandler::Get(const std::string &filename) const {
     std::ifstream stream;
 
     stream.open(filename);
@@ -70,12 +70,12 @@ namespace server::file_handler{
 
   }//Get
 
-  bool file_handler::MyFileHandler::UpDir() {
+  bool MyFileHandler::UpDir() {
     if(chdir("..")==1) return false;
     else return true;
   }//UpDir
 
-  std::vector<std::string> file_handler::MyFileHandler::List() const {
+  std::vector<std::string> MyFileHandler::List() const {
     std::vector<std::string> list;
     for (const auto & file : std::filesystem::directory_iterator(GetCurrentDir())) {
       std::string str(file.path());
@@ -87,7 +87,7 @@ namespace server::file_handler{
 
   }//List
 
-  std::string file_handler::MyFileHandler::GetCurrentDir() const {
+  std::string MyFileHandler::GetCurrentDir() const {
     return std::filesystem::current_path();
   }//GetCurrentDir
 
--- a/project1/server/file_handler/my_file_handler.cpp
+++ b/project1/server/file_handler/my_file_handler.cpp
@@ -1,74 +1,73 @@
 #include "my_file_handler.h"
-#include <bits/stdc++.h>
 
-namespace server::file_handler{
-  bool MyFileHandler::Delete(const std::string &filename) {
+#include <filesystem>
+namespace server::file_handler {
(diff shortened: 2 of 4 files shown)


project1/server/file_handler/CMakeLists.txt
```diff
@@ -1,2 +1,2 @@
-add_library(file_handler file_handler.h)
+add_library(file_handler file_handler.h my_file_handler.h my_file_handler.cpp)
```
djpetti: You can go ahead and remove the `.h` files from this directive. When you have a `.cpp` file, CMake will automatically find any `.h` files that it includes.

project1/server/file_handler/my_file_handler.cpp
```diff
@@ -0,0 +1,98 @@
+//
```
djpetti: I'd remove this.

project1/server/file_handler/my_file_handler.cpp
```diff
@@ -0,0 +1,98 @@
...
+// Created by jake on 1/25/21.
+//
+
+#include "my_file_handler.h"
+#include <bits/stdc++.h>
+
+namespace server::file_handler{
+  bool file_handler::MyFileHandler::Delete(const std::string &filename) {
```
djpetti: Indentation is kind of screwy in this file. If you do `Ctrl-Alt-L` in CLion, it should format it for you.

project1/server/file_handler/my_file_handler.cpp
```diff
@@ -0,0 +1,98 @@
+//
+// Created by jake on 1/25/21.
+//
+
+#include "my_file_handler.h"
+#include <bits/stdc++.h>
+
+namespace server::file_handler{
```
djpetti: This line should make the `file_handler::` prefix on all the function definitions redundant...

project1/server/file_handler/my_file_handler.cpp
```diff
@@ -0,0 +1,98 @@
+//
+// Created by jake on 1/25/21.
+//
+
+#include "my_file_handler.h"
+#include <bits/stdc++.h>
```
djpetti: I think this is an internal part of STL. Don't include it directly. I would individually include all the STL components that you need here.

project1/server/file_handler/my_file_handler.cpp
```diff
@@ -0,0 +1,98 @@
...
+namespace server::file_handler{
+  bool file_handler::MyFileHandler::Delete(const std::string &filename) {
+
+    //convert filename to char*
+    int n = filename.length();
+    char fileName[n+1];
+    strcpy(fileName, filename.c_str());
+
+    //delete file
+    return remove(fileName);
```
djpetti: Is there a reason you can't do `remove(filename.c_str())` directly here?

project1/server/file_handler/my_file_handler.cpp
```diff
@@ -0,0 +1,98 @@
...
+
+#include "my_file_handler.h"
+#include <bits/stdc++.h>
+
+namespace server::file_handler{
+  bool file_handler::MyFileHandler::Delete(const std::string &filename) {
+
+    //convert filename to char*
+    int n = filename.length();
+    char fileName[n+1];
```
djpetti: Nitpick: Let's be consistent on variable naming. The [Google C++ Style Guide](https://google.github.io/styleguide/cppguide.html#Variable_Names) requires lowercase names separated with underscores, as opposed to camelCase.

project1/server/file_handler/my_file_handler.cpp
djpetti: I think there is a subtle bug here. It looks like `remove` [returns 0 upon success](https://en.cppreference.com/w/cpp/io/c/remove) which will decay to `false` since this function returns a bool. That's the opposite of what you want.

Incidentally, I think the compiler can be set to flag implicit conversions like this. I should enable `-Werror` and `-Wall` the next time I commit...

Alternatively, you could use the `std::filesystem` version of [remove](https://en.cppreference.com/w/cpp/filesystem/remove) here, which does return a bool. That might be a good option, considering that you already use this library below.

project1/server/file_handler/my_file_handler.cpp
```diff
@@ -0,0 +1,98 @@
...
+    return remove(fileName);
+  }//Delete
+  bool file_handler::MyFileHandler::Put(const std::string &filename, const std::vector<uint8_t> &contents) {
+
+    //convert contents to string format
+    std::string str_contents(contents.begin(), contents.end());
+
+    //open file
+    try {this->os.open(filename);}
+    catch (std::exception &e) {
```
djpetti: Exceptions in C++ are [not well-liked](https://google.github.io/styleguide/cppguide.html#Exceptions). It's also probably okay to ignore them here, because opening a file in write mode should really not fail except in weird cases where you'd probably want to abort anyway.

project1/server/file_handler/my_file_handler.cpp
```diff
@@ -0,0 +1,98 @@
...
+      this->os.close();
+      return false;
+    }
+
+    //write contents to file
+    try{this->os << str_contents;}
```
djpetti: It's probably okay not to catch exceptions here. This should not throw them except in truly strange conditions that would warrant the program aborting anyway.

project1/server/file_handler/my_file_handler.cpp
```diff
@@ -0,0 +1,98 @@
...
+    this->os.close();
+    return true;
+  }//Put
+  bool file_handler::MyFileHandler::MakeDir(const std::string &name) {
+    //convert directory name to char*
+    int n = name.length();
+    char dir[n+1];
+    strcpy(dir, name.c_str());
+
+    if (mkdir(dir,0777)==-1) return false;
```
djpetti: Again, any reason for not just doing `mkdir(name.c_str(), 0777)` here?

djpetti: [Don't use](https://google.github.io/styleguide/cppguide.html#Conditionals) `if` statements without curly braces. (Google actually makes some exceptions here, but I would argue that they should be flat-out eliminated.) This can easily lead to hard-to-find bugs.

Incidentally, you can achieve the same effect here in a less verbose and error-prone way:
```c++
return mkdir(dir, 0777) != -1
```

project1/server/file_handler/my_file_handler.cpp
```diff
@@ -0,0 +1,98 @@
...
+    else return true;
+  }//MakeDir
+  bool file_handler::MyFileHandler::ChangeDir(const std::string &sub_folder) {
+
+    //convert subdirectory name to char*
+    int n = sub_folder.length();
+    char dir[n+1];
+    strcpy(dir, sub_folder.c_str());
+
+    if(chdir(dir)==-1) return false;
```
djpetti: Same comments here vis-a-vis the argument to `chdir` and lack of curly braces.

project1/server/file_handler/my_file_handler.cpp
```diff
@@ -0,0 +1,98 @@
...
+    else return true;
+  }//ChangeDir
+  std::vector<uint8_t> file_handler::MyFileHandler::Get(const std::string &filename) const {
+    std::ifstream stream;
+
+    stream.open(filename);
+
+    //get contents of file, convert to byte vector
+    std::string str_contents( (std::istreambuf_iterator<char>(stream) ),
```
djpetti: Have you tried simply doing
```c++
std::string str_contents;
stream >> str_contents;
```
The docs suggest that would work, but I haven't tested it.

project1/server/file_handler/my_file_handler.cpp
```diff
@@ -0,0 +1,98 @@
...
+    char dir[n+1];
+    strcpy(dir, sub_folder.c_str());
+
+    if(chdir(dir)==-1) return false;
+    else return true;
+  }//ChangeDir
+  std::vector<uint8_t> file_handler::MyFileHandler::Get(const std::string &filename) const {
+    std::ifstream stream;
+
+    stream.open(filename);
```
djpetti: Any reason why the `ifstream` is a local variable but the `ostream` is a member variable? Unless there is, I would be consistent.

project1/server/file_handler/my_file_handler.cpp
```diff
@@ -0,0 +1,98 @@
...
+
+    std::vector<uint8_t> vec_contents(str_contents.begin(),str_contents.end());
+
+    stream.close();
+    return vec_contents;
+
+  }//Get
+
+  bool file_handler::MyFileHandler::UpDir() {
+    if(chdir("..")==1) return false;
```
djpetti: Again, this can be a one-liner:
```c++
return chdir("..") != 1;
```

project1/server/file_handler/my_file_handler.cpp
```diff
@@ -0,0 +1,98 @@
...
+    else return true;
+  }//UpDir
+
+  std::vector<std::string> file_handler::MyFileHandler::List() const {
+    std::vector<std::string> list;
+    for (const auto & file : std::filesystem::directory_iterator(GetCurrentDir())) {
+      std::string str(file.path());
+      str.erase(0,GetCurrentDir().length());
```
djpetti: Wait, why are you overwriting all the data you just added to the string?

project1/server/file_handler/my_file_handler.cpp
djpetti: Have you considered using [`std::fileystem::create_directory`](https://en.cppreference.com/w/cpp/filesystem/create_directory) instead?

project1/server/file_handler/my_file_handler.h
```diff
@@ -0,0 +1,50 @@
+//
+// Created by jake on 1/25/21.
```
djpetti: Can we put a proper `@file` block here instead? See other `.h` files for examples.

project1/server/file_handler/my_file_handler.h
```diff
@@ -0,0 +1,50 @@
...
+
+
+/**
+ * @brief This class represents common file handling operations.
+ *
+ */
+namespace server::file_handler{
+class MyFileHandler: public IFileHandler {
+
+ private:
```
djpetti: Nitpick: Put `private` section below `public`.

project1/server/file_handler/my_file_handler.h
```diff
@@ -0,0 +1,50 @@
...
+  std::ofstream os;
+
+ public:
+
+  std::vector<uint8_t> Get(const std::string& filename) const;
```
djpetti: Nitpick: Specify [`final`](https://en.cppreference.com/w/cpp/language/final) for these declarations.

project1/server/file_handler/my_file_handler.h
```diff
@@ -0,0 +1,50 @@
...
+
+  bool UpDir();
+
+  bool MakeDir(const std::string& name);
+
+  std::string GetCurrentDir() const;
+
+
+};
+} //namespace server
```
djpetti: `// namespace server::file_handler`

djpetti (changes requested):
I have some suggestions on how this can be improved. Nothing that's too much work, I hope...

project1/server/file_handler/my_file_handler.cpp
jake-chandler: Yeah, looking at the docs for c_str(), the way you suggested it definitely more efficient. I didn't realize the c_str() function returned a const * char.

project1/server/file_handler/my_file_handler.cpp
jake-chandler: Yes, actually because this Get() is a const member function, I am not able to reference any member variables. It might be worth it to remove the const from Get()

project1/server/file_handler/my_file_handler.cpp
jake-chandler: So, the `str(file.path())` here returns a string that contains the entire path to the file. Since list is supposed to return just the filenames, not include the entire path as well, I erase the current working directory substring, leaving only the filename.

Actually, looking over this again made me realize it needs to erase from `str.erase(0,GetCurrentDir().length()+1)`, or a '/'  will be at the front of each filename.

project1/server/file_handler/my_file_handler.cpp
jake-chandler: Have not tried this. Will look into it.

project1/server/file_handler/my_file_handler.cpp
jake-chandler: mkdir() was the first to come to mind, I see no reason to change it. Unless you have other input?

project1/server/file_handler/my_file_handler.cpp
jake-chandler: Not entirely sure which specific STL components are needed here. I'll look into it.

jake-chandler:
Apologies for the commit spam, at first I was committing after fixing each issue you pointed out. Regardless, I've fixed most of them.

project1/server/file_handler/my_file_handler.cpp
djpetti: The only reason I suggested the alternative is because you are using `std::filesystem` functions in other places and consistency never goes unappreciated. It's your choice though.

project1/server/file_handler/my_file_handler.cpp
djpetti: Documentation for individual functions should say which headers you need to include.

project1/server/file_handler/my_file_handler.cpp
```diff
@@ -1,71 +1,47 @@
...
-
-    //convert filename to char*
-    int n = filename.length();
-    char fileName[n+1];
-    strcpy(fileName, filename.c_str());
+  bool MyFileHandler::Delete(const std::string &filename) {
 
-    //delete file
-    return remove(fileName);
+    return remove(filename.c_str());
```
djpetti: Do you still have an issue with return values here?

project1/server/file_handler/my_file_handler.cpp
```diff
@@ -1,71 +1,47 @@
...
-    //convert directory name to char*
-    int n = name.length();
-    char dir[n+1];
-    strcpy(dir, name.c_str());
-
-    if (mkdir(dir,0777)==-1) return false;
-    else return true;
+  bool MyFileHandler::MakeDir(const std::string &name) {
+    
+    return mkdir(name.c_str(),0777)==-1;
```
djpetti: Um... should this be `!= -1`? Or maybe `== 0` would be more readable?

project1/server/file_handler/my_file_handler.cpp
djpetti: Right, I thought you brought this up on Discord and I said that the `const` was an uninformed assumption on my part, and it was fine to remove it if necessary. In this case, I'd prefer either using local `fstream`s everywhere, or removing the `const` and using member `fstream`s everywhere. In fact, I think I slightly prefer the former, as I don't really see that having the member variables buys you anything in this case. (Maybe a small amount of avoided memory allocations? Nothing I would lose sleep over though...)

project1/server/file_handler/my_file_handler.h
```diff
@@ -22,12 +22,9 @@
...
 
- private:
-  std::ofstream os;
-
  public:
 
-  std::vector<uint8_t> Get(const std::string& filename) const;
+  final std::vector<uint8_t> Get(const std::string& filename) const;
 
   bool Put(const std::string& filename,
```
djpetti: Go ahead and make these other public methods `final` as well.

djpetti (commented):
Hopefully I didn't review this too early... It looks like you're still making some changes?

djpetti:
> Apologies for the commit spam, at first I was committing after fixing each issue you pointed out. Regardless, I've fixed most of them.

This is okay, if I'm feeling really anal, we can squash them before merging.

project1/server/file_handler/my_file_handler.cpp
jake-chandler: Yeah, `== 0 `, it should be.

project1/server/file_handler/my_file_handler.cpp
jake-chandler: Yeah, I like removing the fstream member variable altogether. Looks like it was only used in one other function anyways.

project1/server/file_handler/my_file_handler.cpp
jake-chandler: yeah, I like consistency.

jake-chandler:
Alright, I think I've covered all of the change requests. It looks much cleaner than before. I believe you're ready to merge.

project1/server/file_handler/my_file_handler.cpp
```diff
@@ -1,73 +1,73 @@
...
+
+#include <filesystem>
+namespace server::file_handler {
+bool MyFileHandler::Delete(const std::string &filename) {
+  return remove(filename.c_str());
+
+}  // Delete
+bool MyFileHandler::Put(const std::string &filename,
+                        const std::vector<uint8_t> &contents) {
+  std::ofstream stream;
```
djpetti, line 11: (minor) Can you just to `std::ofstream stream(filename)`?

djpetti (approved):
Minor comments here. Once all are resolved, this is good to merge.

project1/server/file_handler/my_file_handler.cpp
djpetti: Bump

project1/server/file_handler/my_file_handler.cpp
djpetti: Bump

project1/server/file_handler/my_file_handler.cpp
jake-chandler: I see! Just fixed this.

project1/server/file_handler/my_file_handler.cpp
jake-chandler: got it.

project1/server/file_handler/my_file_handler.cpp
jake-chandler: I've tested this a couple times and I keep getting errors. Will keep this the same for now, but I'm still going to look into it more.

PR #3, opened 2021-01-30
Make wire protocol actually work.

djpetti:
This was done with the help of some unit tests. These tests are located in `project1/wire_protocol/tests`, and should also serve as good examples for anyone trying to use the wire protocol.

Running the unit tests can be done easily using CLion, just by adding a new GTest configuration and selecting `test_wire_protocol` as the executable.


jake-chandler (approved):
I looked through wire_protocol.h and wire_protocol.cpp in much more detail, but don't have any suggestions about how to improve. I also didn't catch any obvious/immediate errors in any of your parsing logic. That being said, I think it's ready to merge.

PR #4, opened 2021-02-06
Add initial server code.

djpetti:
This remains untested, since the client does not exist yet.

Also note that the implementation is very bare-bones. Error handling, for instance, is almost non-existent; if you ask it to read a file that doesn't exist it will simply disconnect you without saying anything.


jake-chandler (approved):
Looks good. Your Agent class is clever, nice job. Will be mimicking your basic logic for sending/receiving data through the socket.

PR #5, opened 2021-02-08
first implementation of FTP Client

jake-chandler:
This is not completely finished. I am concerned that the functions  `MakeAddress` and `SetUpSocket` could be implemented incorrectly. I would like a second pair of eyes to examine these. 

I do believe that my processes for creating/sending requests, and receiving/interpreting responses are implemented correctly. 
I also believe that the shell logic is sound.

There is definitely some redundancy in the `InputParser` class (which leads to redundancy in the `client` class) that I pointed out in some of my comments. I think this redundancy could be fixed with Templates.

I know for sure there is some nit-picky stuff that needs to be tended to in my code. If it doesn't invalidate functionality, it will be resolved if time permits. Although I don't think there should be any formatting issues because I was consistently using cntrl-alt-L in CLion.

I believe the next immediate steps are to make sure my socket connection process is sound, then perhaps we can perform testing together over a discord call.

### DIFF
--- a/project1/client/CMakeLists.txt
+++ b/project1/client/CMakeLists.txt
@@ -1,0 +1,3 @@
+add_subdirectory(input_parser)
+add_library(client input_parser/input_parser.h client.cpp client.h client_main.cpp)
+target_link_libraries(client wire_protocol)
--- a/project1/client/CMakeLists.txt
+++ b/project1/client/CMakeLists.txt
@@ -1,3 +1,7 @@
 add_subdirectory(input_parser)
-add_library(client input_parser/input_parser.h client.cpp client.h client_main.cpp)
-target_link_libraries(client wire_protocol)
+add_library(client input_parser/input_parser.h client.cpp client.h)
+target_link_libraries(client wire_protocol file_handler input_parser)
+add_executable(client_exe client_main.cpp)
+target_link_libraries(client_exe client)
+
+
--- a/project1/client/client.cpp
+++ b/project1/client/client.cpp
@@ -1,5 +1,183 @@
-//
-// Created by jake on 2/7/21.
-//
-
 #include "client.h"
+#include <sys/socket.h>
+#include <netinet/in.h>
+#include <iostream>
+
+
+namespace client {
+    using client::input_parser::InputParser;
+
+    /**
+     * @brief helper function for making a socket address structure
+     *
+     * @note could contain some fundamental errors...
+     */
+    struct sockaddr_in MakeAddress(uint16_t port) {
+        struct sockaddr_in address{};
+        address.sin_family = AF_INET;
+        address.sin_addr.s_addr = INADDR_ANY;
+        address.sin_port = htons(port);
+
+        return address;
+    }
+
+    /**
+     * @brief connects to socket
+     * @param address socket address structure
+     * @param name FTP Server IP Address
+     * @return socket fd on success, -1 on failure
+     *
+     * @note could contain some fundamental errors...
+     */
+    int SetUpSocket(const struct sockaddr_in &address, std::string &name) {
+        int sock = 0;
+
+        if ((sock = socket(AF_INET, SOCK_STREAM, 0)) < 0) {
+            perror("Socket creation error");
+            return -1;
+        }
+
+        if (inet_pton(AF_INET, name.c_str(), (struct sockaddr *) &address.sin_addr) <= 0) {
+            perror("Invalid Address or Address not supported");
+            return -1;
+        }
+
+        if (connect(sock, (struct sockaddr *) &address, sizeof(address)) < 0) {
+            perror("Connection Failed");
+            return -1;
+        }
+
+
+        return sock;
+    }
+
+    bool Client::Connect(std::string name, uint16_t port) {
+        client_fd = SetUpSocket(MakeAddress(port), name);
+        connected = true;
+        return connected;
+    }
+
+    bool Client::SendReq() {
+        if (!connected) {
+            perror("Cannot Send Request... no longer connected to server");
+        }
+        if (send(client_fd, outgoing_msg_buf_.data(), outgoing_msg_buf_.size(), 0) < 0) {
+            perror("Failed to send request.");
+            return false;
+        }
+        return true;
+    }
+
+    void Client::HandleResponse() {
+        output = "";
+
+        //not sure if this is bad or not.. CLion yelled at me for not initializing.
+        ftp_messages::Response *msg = nullptr;
+
+        parser_.GetMessage(msg);
+
+        //determine the type of response
+        if (msg->has_get()) {
+
+            //I'm not 100% about this casting but I think it's correct.
+            ftp_messages::GetResponse *gResponse = (ftp_messages::GetResponse *) msg;
+
+            //no, I don't use the FileHandler class I made. will change eventually
+            std::ofstream new_file;
+            new_file.open(ip->GetFilename());
+            new_file << gResponse->file_contents();
+        } else if (msg->has_list()) {
+
+            //I'm not 100% about this casting but I think it's correct.
+            ftp_messages::ListResponse *lResponse = (ftp_messages::ListResponse *) msg;
+
+            //append filenames to output string separated by whitespace
+            for (int i = 0; i < lResponse->filenames_size(); i++) {
+                output.append(lResponse->filenames(i));
+                if (!(i == lResponse->filenames_size() - 1)) {
+                    output.append(" ");
+                }
+            }
+        } else if (msg->has_pwd()) {
+
+            //I'm not 100% about this casting but I think it's correct.
+            ftp_messages::PwdResponse *pwdResponse = (ftp_messages::PwdResponse *) msg;
+            output = pwdResponse->dir_name();
+        }
+
+        //make sure outputting is necessary
+        if (output.compare("")) {
+            //does nothing currently
+            Output();
+        }
+
+    }
+
+    void Client::FtpShell() {
+
+        do {
+            //reset input string
+            std::string input = "";
+
+            //will have to pay attention to formatting here when implementing Output()
+            std::cout << "\nmyftp> ";
+
+            //get input
+            std::cin >> input;
+
+            //determine command
+            ip = new InputParser(input);
+            InputParser::ReqType req = ip->GetReqType();
+
+            //create & serialize request message for determined command
+            switch (req) {
+                case InputParser::GETF:
+                    g = ip->CreateGetReq();
+                    SerializeMessage(g);
+                    break;
+                case InputParser::PUTF:
+                    p = ip->CreatePutReq();
+                    SerializeMessage(p);
+                    break;
+                case InputParser::DEL:
+                    del = ip->CreateDelReq();
+                    SerializeMessage(del);
+                    break;
+                case InputParser::LS:
+                    ls = ip->CreateListReq();
+                    SerializeMessage(ls);
+                    break;
+                case InputParser::CD:
+                    cd = ip->CreateCDReq();
+                    SerializeMessage(cd);
+                    break;
+                case InputParser::MKDIR:
+                    mkdir = ip->CreateMkdirReq();
+                    SerializeMessage(mkdir);
+                    break;
+                case InputParser::PWD:
+                    pwd = ip->CreatePwdReq();
+                    SerializeMessage(pwd);
+                    break;
+                case InputParser::QUIT:
+                    q = ip->CreateQuitReq();
+                    SerializeMessage(q);
+
+                    //close socket connection
+                    close(client_fd);
+                    connected = false;
+
+                    //should skip this iteration of the loop,
+                    //avoiding the functions called below.
+                    //basically this should exit the do-while
+                    continue;
+            }
+            SendReq();
+            WaitForResponse();
+            HandleResponse();
+
+        } while (connected);
+
+
+    }
+}
--- a/project1/client/client.cpp
+++ b/project1/client/client.cpp
@@ -1,0 +1,5 @@
+//
+// Created by jake on 2/7/21.
+//
+
+#include "client.h"
--- a/project1/client/client.cpp
+++ b/project1/client/client.cpp
@@ -4,13 +4,12 @@
 #include <iostream>
 
 
-namespace client {
-    using client::input_parser::InputParser;
+namespace {
 
     /**
      * @brief helper function for making a socket address structure
      *
-     * @note could contain some fundamental errors...
+     * @param port FTP server Port #
      */
     struct sockaddr_in MakeAddress(uint16_t port) {
         struct sockaddr_in address{};
@@ -24,12 +23,11 @@ namespace client {
     /**
      * @brief connects to socket
      * @param address socket address structure
-     * @param name FTP Server IP Address
+     * @param hostname FTP server IP address
      * @return socket fd on success, -1 on failure
      *
-     * @note could contain some fundamental errors...
      */
-    int SetUpSocket(const struct sockaddr_in &address, std::string &name) {
+    int SetUpSocket(const struct sockaddr_in &address, const std::string &hostname) {
         int sock = 0;
 
         if ((sock = socket(AF_INET, SOCK_STREAM, 0)) < 0) {
@@ -37,7 +35,7 @@ namespace client {
             return -1;
         }
 
-        if (inet_pton(AF_INET, name.c_str(), (struct sockaddr *) &address.sin_addr) <= 0) {
+        if (inet_pton(AF_INET, hostname.c_str(), (struct sockaddr *) &address.sin_addr) <= 0) {
             perror("Invalid Address or Address not supported");
             return -1;
         }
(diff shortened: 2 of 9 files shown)


jake-chandler:
also not sure about this merge conflict

project1/client/client.cpp
```diff
@@ -0,0 +1,183 @@
...
+
+namespace client {
+    using client::input_parser::InputParser;
+
+    /**
+     * @brief helper function for making a socket address structure
+     *
+     * @note could contain some fundamental errors...
+     */
+    struct sockaddr_in MakeAddress(uint16_t port) {
```
djpetti, line 14: (minor) These two functions should probably be in an anonymous namespace.

project1/client/client.cpp
```diff
@@ -0,0 +1,183 @@
...
+
+    bool Client::Connect(std::string name, uint16_t port) {
+        client_fd = SetUpSocket(MakeAddress(port), name);
+        connected = true;
+        return connected;
+    }
+
+    bool Client::SendReq() {
+        if (!connected) {
+            perror("Cannot Send Request... no longer connected to server");
```
djpetti, line 65: Did you mean to return early at the end of this `if`` statement?

project1/client/client.cpp
```diff
@@ -0,0 +1,183 @@
...
+            perror("Failed to send request.");
+            return false;
+        }
+        return true;
+    }
+
+    void Client::HandleResponse() {
+        output = "";
+
+        //not sure if this is bad or not.. CLion yelled at me for not initializing.
```
djpetti: I think what you're doing is right.

Also, a note for the future: these sorts of thoughts are better expressed as review comments on your own PR (there's no rule against doing that) than comments in the code.

project1/client/client.cpp
```diff
@@ -0,0 +1,183 @@
...
+        ftp_messages::Response *msg = nullptr;
+
+        parser_.GetMessage(msg);
+
+        //determine the type of response
+        if (msg->has_get()) {
+
+            //I'm not 100% about this casting but I think it's correct.
+            ftp_messages::GetResponse *gResponse = (ftp_messages::GetResponse *) msg;
```
djpetti: Uh... not quite. This is going to be undefined behavior. I think what you're looking for is `const ftp_messages::GetResponse *gResponse = msg->get();` [This documentation](https://developers.google.com/protocol-buffers/docs/reference/cpp-generated#oneof-string-fields) should clear up how this works.

Also, some more notes for future reference:
- Don't use C-style casts. (I'm surprised CLion doesn't warn you about that). `static_cast`, `dynamic_cast`, `reinterpret_cast` et. al. are your friends.
-  Local variable naming convention dictates that this should be called `g_response`.
- This is a situation when using the `auto` keyword might be in your best interest.

project1/client/client.cpp
```diff
@@ -0,0 +1,183 @@
...
+
+            //no, I don't use the FileHandler class I made. will change eventually
```
djpetti: Again, this is not really the way I'd express this kind of thought. If you really are serious about doing this, it should be a TODO, like `// TODO (jakechandler) Change this to use FileHandler instead.` Although, honestly, this is, once again, probably best expressed as a self-review comment.

project1/client/client.cpp
```diff
@@ -0,0 +1,183 @@
...
+            std::ofstream new_file;
+            new_file.open(ip->GetFilename());
+            new_file << gResponse->file_contents();
+        } else if (msg->has_list()) {
+
+            //I'm not 100% about this casting but I think it's correct.
+            ftp_messages::ListResponse *lResponse = (ftp_messages::ListResponse *) msg;
```
djpetti: ditto

project1/client/client.cpp
```diff
@@ -0,0 +1,183 @@
...
+            for (int i = 0; i < lResponse->filenames_size(); i++) {
+                output.append(lResponse->filenames(i));
+                if (!(i == lResponse->filenames_size() - 1)) {
+                    output.append(" ");
+                }
+            }
+        } else if (msg->has_pwd()) {
+
+            //I'm not 100% about this casting but I think it's correct.
+            ftp_messages::PwdResponse *pwdResponse = (ftp_messages::PwdResponse *) msg;
```
djpetti: Same here.

(Probably don't need to flag all of these, but there's a good chance that as written this would segfault or do something weird, so just want to make sure you get them all.)

project1/client/client.h
```diff
@@ -0,0 +1,144 @@
...
+
+#include <cstdint>
+#include "../wire_protocol/wire_protocol.h"
+#include "../server/file_handler/my_file_handler.h"
+#include "ftp_messages.pb.h"
+#include "input_parser/input_parser.h"
+
+namespace client {
+
+    using ftp_messages::Request;
```
djpetti: I think I mentioned this in passing, but I usually don't use `using` in `.h` files, because it can lead to unexpected behavior for people `#including` them directly (or even worse, transitively) down the line.

project1/client/client.cpp
```diff
@@ -0,0 +1,183 @@
+#include "client.h"
+#include <sys/socket.h>
+#include <netinet/in.h>
+#include <iostream>
+
+
+namespace client {
```
djpetti: Is CLion responsible for this weird formatting where the namespace contents are indented? That's kind of odd...

project1/client/client.cpp
```diff
@@ -0,0 +1,183 @@
...
+                    SerializeMessage(q);
+
+                    //close socket connection
+                    close(client_fd);
+                    connected = false;
+
+                    //should skip this iteration of the loop,
+                    //avoiding the functions called below.
+                    //basically this should exit the do-while
+                    continue;
```
djpetti, line 207: Wait, but won't this force the loop to run one more time, waiting for a new command from the user? Perhaps you want this to be a `while` loop, not a `do while` loop?

project1/client/client.cpp
```diff
@@ -0,0 +1,183 @@
...
+            std::string input = "";
+
+            //will have to pay attention to formatting here when implementing Output()
+            std::cout << "\nmyftp> ";
+
+            //get input
+            std::cin >> input;
+
+            //determine command
+            ip = new InputParser(input);
```
djpetti: Now, this is an excellent example of why *not* to use `new`. You see, there's no corresponding `delete` here, so this leaks memory. Nowadays, if you want to use dynamic allocation in C++, I would strongly recommend using `std::unique_ptr` or `std::shared_ptr`, depending on the use-case. (I find that `unique_ptr` is appropriate 90% of the time.)

Or, you could just allocate it on the stack. Is there some reason you can't do that which I'm not noticing?

project1/client/client.h
```diff
@@ -0,0 +1,144 @@
...
+    using ftp_messages::QuitRequest;
+    using ftp_messages::MakeDirRequest;
+
+    using wire_protocol::MessageParser;
+
+    /**
+     * @brief FTP Client. Contains logic for connecting to server, sending requests, outputting responses,
+     * and shell continuity.
+     *
+     * @note might rename to FTPClient or a better name
```
djpetti: Again, probably wouldn't make this a code comment. This can be a `TODO` or a self-review comment, depending on its urgency.

project1/client/client.h
```diff
@@ -0,0 +1,144 @@
...
+        void FtpShell();
+
+        /**
+         * @brief Handles output formatting for responses
+         */
+        void Output();
+
+        /**
+         *
+         * @param name the ip address of the FTP server
```
djpetti: (nitpick) You *may* consider calling this parameter `hostname`, since just `name` is a little... subtle.

project1/client/client.h
```diff
@@ -0,0 +1,144 @@
...
+         * @param port the port the FTP server is binded to
+         * @return true on success, false on failure
+         */
+        bool Connect(std::string name, uint16_t port);
```
djpetti: (minor) Make this a `const std::string&` instead.

In general, any "complicated" type should be passed by constant reference when you can get away with it.

project1/client/client.h
```diff
@@ -0,0 +1,144 @@
...
+         * @brief serializes the message and store results to be used by SendReq()
+         * @tparam T , expected to be of type ftp::messages
+         * @param req the request type
+         * @return true on success, false on failure
+         */
+        template<class T>
+        bool SerializeMessage(T &req) {
+            MessageParser<T> parser;
+
+            const bool kSerializedSuccess = wire_protocol::Serialize(req, &outgoing_msg_buf_);
```
djpetti: Okay, you're mixing up serialization and de-serialization. The `MessageParser` class is for deserialization exclusively. If all you're trying to do is serialize the message, all you need is line 64. Luckily, the serialization process is actually pretty simple.

I admire your use of templates and all, but sadly, this whole function is probably unnecessary.

project1/client/client.h
```diff
@@ -0,0 +1,144 @@
...
+         * @brief sends the serialized message via a socket
+         * @return true on success, false on failure
+         */
+        bool SendReq();
+
+        /**
+         * @brief expects a response message from socket, stores message in parser
+         *
+         */
+        void WaitForResponse() {
```
djpetti: Any reason this is living in the `.h` file?

project1/client/client.h
```diff
@@ -0,0 +1,144 @@
...
+            while (!parser_.HasCompleteMessage()) {
+                incoming_msg_buf_.resize(kBufferSize);
+
+                const auto bytes_read =
+                        recv(client_fd, incoming_msg_buf_.data(), kBufferSize, 0);
+                if (bytes_read < 0) {
+                    //reset parser?
```
djpetti: This is an error condition. In practice, when (if) you hit this, it's pretty likely that your entire socket is borked, so resetting the parser and trying again is probably just going to get you stuck in an infinite loop. Probably better to print an error, and then return false (requires making this function return a `bool`, of course).

project1/client/client.h
```diff
@@ -0,0 +1,144 @@
...
+        //response info to be formatted and outputted
+        std::string output;
+
+        //parser for handling messages
+        MessageParser<ftp_messages::Response> parser_;
+
+        //parser for user input
+        client::input_parser::InputParser* ip;
+
+        GetRequest g;
```
djpetti: Must these be class members? It seems like you only use them in one method?

project1/client/client.h
```diff
@@ -0,0 +1,144 @@
...
+
+        //buffer that stores serialized data to be sent to and received from server
+        std::vector<uint8_t> outgoing_msg_buf_{};
+        std::vector<uint8_t> incoming_msg_buf_{};
+
+        //buffer size for client.
+        static constexpr size_t kBufferSize = 4096;
+
+        //client socket fd
+        int client_fd;
```
djpetti: (minor) The [style-guide convention](https://google.github.io/styleguide/cppguide.html#Variable_Names) is to use a trailing underscore for class data members, so they can be differentiated from local variables when reading a .cpp file.

project1/client/client.h
```diff
@@ -0,0 +1,144 @@
...
+        ChangeDirRequest cd;
+        ListRequest ls;
+        PwdRequest pwd;
+        DeleteRequest del;
+        QuitRequest q;
+        MakeDirRequest mkdir;
+        Request r;
+
+    };
+}
```
djpetti, line 89: (minor) `}  // namespace client`

project1/client/client_main.cpp
```diff
@@ -0,0 +1,28 @@
+/**
+ * @brief Entry point for the FTP client.
+ */
+
+
+#include "client.h"
+#include "input_parser/input_parser.h"
+#define PORT 8080
+namespace client {
```
djpetti: `main` can't be in a namespace for this to work. I think the only reason this compiles at all is because you're not actually making an executable. (See my comment above.)

project1/client/CMakeLists.txt
```diff
@@ -0,0 +1,3 @@
+add_subdirectory(input_parser)
+add_library(client input_parser/input_parser.h client.cpp client.h client_main.cpp)
```
djpetti: I don't think you want `client_main.cpp` to be part of this library, right? You probably want to remove that and add something like
```
add_executable(client_exe client_main.cpp)
target_link_libraries(client_main client)
```

Although, for consistency with the server, I'd call the library `client_lib` or something, and the executable `client`.

project1/client/client_main.cpp
```diff
@@ -0,0 +1,28 @@
...
+
+
+    int main(int argc, const char **argv) {
+        Client c;
```
djpetti: (minor) Good practice is to check the number of arguments and print a help message if it's wrong. Check out what I did in the equivalent server code.

project1/client/input_parser/input_parser.cpp
```diff
@@ -0,0 +1,157 @@
...
+    using ftp_messages::PutRequest;
+    using ftp_messages::ChangeDirRequest;
+    using ftp_messages::ListRequest;
+    using ftp_messages::PwdRequest;
+    using ftp_messages::DeleteRequest;
+    using ftp_messages::QuitRequest;
+    using ftp_messages::MakeDirRequest;
+
+    InputParser::InputParser(std::string &cmd) {
+        this->fn = "";
```
djpetti, line 15: (minor) These variable names aren't very descriptive. Also, is there a reason you have to prefix with `this->`?

project1/client/input_parser/input_parser.h
```diff
@@ -0,0 +1,105 @@
...
+#include <string>
+#include <vector>
+
+namespace client::input_parser {
+
+    /**
+     * @brief Assists in mapping input commands for FTP client to corresponding
+     * protobuf requests.
+     *
+     * @note There's definitely some redundancy here. Will likely change
```
djpetti, line 18: I'm not sure how much you can slim this down, considering that each request type requires somewhat unique code to populate it. If you really want to pretify this code, we can talk about it at one of our meetings; I have a few tricks up my sleeve for dealing with situations like this.

project1/client/input_parser/input_parser.h
```diff
@@ -0,0 +1,105 @@
...
+     * this class to have templated CreateReq functions in project 2.
+     */
+    class InputParser {
+    public:
+        enum ReqType {GETF, PUTF, DEL, LS, CD, MKDIR, PWD, QUIT};
```
djpetti, line 23: Use `enum class` here instead. Strongly-typed enums in C++ [are generally preferable to C enums](https://stackoverflow.com/a/12581154/1516662).

project1/client/input_parser/input_parser.h
```diff
@@ -0,0 +1,105 @@
...
+
+        /**
+         * @brief Initializes relevant info about this input command
+         * @param cmd the user input
+         * @note assumes correct syntax
+         */
+        InputParser(std::string& cmd);
+        ~InputParser();
```
djpetti, line 31: Did you actually define a destructor and I'm just missing it? Otherwise, I'm surprised this compiles. For this class, it's probably fine to remove this line and just use the default destructor.

project1/client/input_parser/input_parser.h
```diff
@@ -0,0 +1,105 @@
...
+        /**
+         * @brief getter for filename
+         * @return the filename
+         */
+        std::string GetFilename();
+        /**
+         * creates a get request
+         * @return the request
+         */
+        ftp_messages::GetRequest CreateGetReq();
```
djpetti, line 49: One minor thing you can do is take advantage of overloading like I did in my `Agent` class. That way, you don't have to explicitly remember to call the right version of this function.

project1/client/input_parser/input_parser.h
```diff
@@ -0,0 +1,105 @@
...
+        /**
+         * creates a quit request
+         * @return the request
+         */
+        ftp_messages::QuitRequest CreateQuitReq();
+
+    private:
+
+
+        const std::string commands[8] = {"get", "put", "delete", "ls", "cd", "mkdir", "pwd", "quit"};\
```
djpetti, line 96: (minor) This should be called `kCommands`. Also, you can probably get away with not specifying the array length in this case? Also, I'm surprised this compiles without it also being `static`...

djpetti (changes requested):
Much of the comments on here are cosmetic, but there are a few issues which will prevent this code from working.

project1/client/client.cpp
jake-chandler: Yes. I don't really understand why CLion formats the namespace contents indented. I thought they weren't supposed to be so I was confused.

djpetti: This might be an issue with your `clang-format` configuration. It's probably easier to look into it at our next meeting than worry about it now.

project1/client/client.h
jake-chandler: agreed... lol very subtle

project1/client/client.h
jake-chandler: Okay that makes a lot more sense I was a bit confused when I first wrote this...
I'll remove the function. Even though yes i was proud of using templates.

project1/client/client.h
jake-chandler: At first it was a templated function. I realized it didn't need to be and forgot to move it.

project1/client/client.h
jake-chandler: yeah, going to make this a bool,

PR #6, opened 2021-02-22
Add an interface for the thread pool.

djpetti:


project1/thread_pool/task.h
```diff
@@ -0,0 +1,60 @@
...
+
+namespace thread_pool {
+
+/**
+ * @brief Represents a task that can be run in a tread pool.
+ */
+class Task {
+ public:
+  /// Represents an opaque identifier for a task.
+  using Handle = uint32_t;
```
jake-chandler, line 15: You're mapping the keyword `Handle` to the datatype `uint32_t` ? 

Is this for readability purposes only?

project1/thread_pool/task.h
```diff
@@ -0,0 +1,60 @@
...
+   */
+  virtual Status SetUp() = 0;
+
+  /**
+   * @brief Runs one iteration of the task main loop.
+   * @note The task cannot be cancelled in the middle of this function. However,
+   *    you should assume that cancellation is possible between calls.
+   * @return The updated status for this task.
+   */
+  virtual Status RunAtomic() = 0;
```
jake-chandler, line 47: I'm not sure what you mean by `task main loop`? Where is the `task main loop` stored in this `Task` class? This might be a poorly worded question, but it seems like the `Task` class has lots of information about the task but the actual task itself.

project1/thread_pool/task.h
```diff
@@ -0,0 +1,60 @@
...
+  virtual void Cleanup() = 0;
+
+  /**
+   * @brief Gets a unique ID for this task.
+   */
+  Handle GetId();
+
+ private:
+  /// Static counter we use for generating task IDs.
+  static uint32_t current_id_;
```
jake-chandler: The idea would be to increment this variable when new tasks are made and then decrement this variable when tasks fail or are completed?

project1/thread_pool/thread_pool_interface.h
```diff
@@ -0,0 +1,41 @@
...
+#include <memory>
+
+#include "task.h"
+
+namespace thread_pool {
+
+/**
+ * @brief Standardized interface for thread pools.
+ */
+class IThreadPool {
```
jake-chandler, line 14: I don't see a `RunTask` function or equivalent. I'm not experienced with threads, so I'm not exactly sure how a task will run.

project1/thread_pool/task.h
```diff
@@ -0,0 +1,60 @@
+#ifndef PROJECT1_TASK_H
+#define PROJECT1_TASK_H
+
+#include <cstdint>
+
+namespace thread_pool {
+
+/**
+ * @brief Represents a task that can be run in a tread pool.
```
jake-chandler: typo here `tread pool`.

project1/thread_pool/task.h
```diff
@@ -0,0 +1,60 @@
...
+    /// Task is running normally.
+    RUNNING,
+    /// Task has completed normally.
+    DONE,
+    /// Task has failed.
+    FAILED,
+  };
+
+  /**
+   * @brief Performs any necessary one-time setup when the task starts.
```
jake-chandler, line 36: What if certain tasks require different one-time setups? I'm slightly confused because different tasks could have different set up requirements? Maybe I'm thinking about this the wrong way.

jake-chandler (changes requested):
I mostly just have some high level questions about this. There's really only one thing that needs to be changed certainly and it's merely a typo.

project1/thread_pool/task.h
djpetti: Yes. Also because `Handle` is really just an opaque identifier and it doesn't really matter what type it is, as long as it's hashable.

project1/thread_pool/task.h
djpetti: After reading over this and some of your comments below, I think there might be some confusion on how this is intended to be used. I think this will become much more clear once I submit the actual implementation and unit tests.

The `Task` class is not meant to be used directly. New tasks are meant to be created by subclassing `Task` and implementing `RunAtomic` so that it does whatever you want your task to do. The terminology stems from the notion that most tasks are expected to consist of some loop (i.e. a "main loop") that gets run repeatedly. For instance, you might have a task that listens on a socket, handles a command that it received, and then goes back to listening on the socket for the next command.

I recognize that this implementation strategy is a little confusing compared to just having a `Run` method that runs the entire task in a monolithic block. It is done this way in order to implement task cancellation without having to have any (explicit) buy-in from specific task implementations.

project1/thread_pool/task.h
djpetti: It's actually never decremented, as doing so would allow multiple running tasks to end up with the same ID, but yeah. All this logic is done automatically, by the way, so people implementing `Task` don't have to worry about this (and it's private anyway, so subclasses couldn't touch it even if they wanted to).

project1/thread_pool/thread_pool_interface.h
djpetti: See my comment above about the `RunAtomic` method in `Task`.

In this case, a new task is submitted to the pool by calling `AddTask`. This will automatically run the task as soon as possible.

project1/thread_pool/task.h
djpetti: Whoops

djpetti:
At this point, I've already almost finished the implementation of `ThreadPool`, so I think I'm just going to adopt this same PR for that as well. Reviewers, take note.

project1/thread_pool/task.h
djpetti: A specific one-time setup can be implemented for every `Task` subclass.

project1/thread_pool/thread_pool_interface.h
```diff
@@ -16,26 +17,30 @@ class IThreadPool {
...
+  virtual void AddTask(const std::shared_ptr<Task>& task) = 0;
 
   /**
    * @brief Gets the current status of task.
-   * @param task_handle The handle of the task.
+   * @param task The task..
    * @return The status of the task.
    */
-  virtual Task::Status GetTaskStatus(const Task::Handle& task_handle) = 0;
+  virtual Task::Status GetTaskStatus(const std::shared_ptr<Task>& task) = 0;
```
jake-chandler, line 29: Why the change here? Just curious.

jake-chandler (approved):
Daniel, you were right about the Unit Testing clearing up the intended utility of the `Task` class. Now from my current understanding, I'll be making `Task` sub-classes for each task the `Server` will perform, while using the `ThreadPool` to manage these tasks.  Now I'll review the server code more and brainstorm what exactly will qualify as a task. 

Thanks for the unit tests.

project1/thread_pool/thread_pool_interface.h
djpetti: In my initial implementation, I was using `std::unique_ptr` for tasks and I actually had the pool take complete ownership of the task. This meant that the main thread had no access to the task after calling `AddTask`. Because of that, all the methods like `GetTaskStatus` had to use the task handle, which could easily be saved, as opposed to the task itself.

However, as I actually tried to use this implementation, I quickly realized that it was annoying. There are, in fact, legitimate reasons why you want the main thread to still have access to submitted tasks. For instance, if you have tasks that produce results, the "correct" way to implement that is to store the result as a member of the `Task` itself and then query the result in the main thread once the task is complete.

Therefore, I transitioned to using `std::shared_ptr` for tasks instead, which allows for both paradigms: If you don't want to keep the task around in the main thread, you don't have to, and the thread pool will automatically take sole ownership of the task, just as before. However, if you want to keep it around, you can do that too, and now the main thread and pool will share ownership.

Once it became possible to hold onto `Task` objects in the main thread, using the task handle for everything became unnecessary, and the API was much more unwieldy than just passing the actual `Task` object. That's why I ultimately changed it.

djpetti:
> Daniel, you were right about the Unit Testing clearing up the intended utility of the `Task` class. Now from my current understanding, I'll be making `Task` sub-classes for each task the `Server` will perform, while using the `ThreadPool` to manage these tasks. Now I'll review the server code more and brainstorm what exactly will qualify as a task.
> 
> Thanks for the unit tests.

You're welcome. I'm glad everything makes more sense now.

project1/thread_pool/thread_pool.h
```diff
@@ -0,0 +1,102 @@
...
+  /// Maps task handles to threads.
+  std::unordered_map<Task::Handle, std::thread> handle_to_thread_{};
+  /// Handles of tasks that should be cancelled.
+  std::unordered_set<Task::Handle> cancelled_tasks_{};
+
+  /**
+   * @brief Indicates that we have a pending task that's ready to run. Also used
+   *    to indicate that the pool is being closed.
+   */
+  std::condition_variable task_pending_;
```
Nickster258, line 70: Two brief thoughts here:
- If tasks are supposed to run at first possible chance with the use of `Atomic`, when would this be used to find out if a task has yet to run, assuming it runs immediately? Does this have something to do with `SetUp`?
- I am not sure why `task_pending_` and `task_done_` are both used to indicate the pool is being closed. Is there a reason why `pool_closing_` isn't also a variable?

project1/thread_pool/thread_pool.h
```diff
@@ -0,0 +1,102 @@
...
+ private:
+  /**
+   * @brief Thread that's in charge of dispatching new tasks to the pool.
+   */
+  void DispatcherThread();
+
+  /**
+   * @brief Thread that's in charge of finalizing completed tasks.
+   */
+  void JoinerThread();
```
Nickster258, line 41: Am I correct in assuming that tasks get set up and completed synchronously by the Pool but run its task asynchronously by itself?

Nickster258 (commented):
I mostly only had a couple questions regarding the thread pool. Other questions I had were answered in @jake-chandler's review.

project1/thread_pool/thread_pool.h
djpetti: Do you mean `RunAtomic`?

At the "first possible chance" is not the same as "immediately" in this case. The task lifecycle actually looks like this:
- `AddTask` is called and the task is registered and added to a runnable queue.
- A special dispatcher thread reads the task from the runnable queue, and creates a new thread to run the task in.
- Within this task-specific thread:
  - The `SetUp` method is called.
  - `RunAtomic` is called repeatedly until it returns a status other than `RUNNING`.
  - The `CleanUp` method is called.
  - The task is added to the joinable queue.
- Another special joiner thread reads the task from the joinable queue, and cleans up all the state associated with the task.

Basically, we need a way for the dispatcher thread to know when we have called `AddTask` and there is a new runnable task. That's what that condition variable is for.

As to your second bullet point, I think there is a slight conceptual misunderstanding here, probably owing to poorly-named STL types. `std::condition_variable`, despite the name, doesn't encapsulate any actual state; it's merely a synchronization primitive. The state that you're checking the condition of actually has to be stored separately. In this case, that state is `should_stop_`. I'm merely notifying on all the condition variables to force all the threads that might be waiting on them to wake up, at which point they'll immediately check `should_stop_`, see that it is true, and dutifully exit.

project1/thread_pool/thread_pool.h
djpetti: Sort of. See my "task lifecycle" explanation in the previous comment.

**TL;DR:** Though some of the task initialization is done synchronously, much of the initialization and all of the finalization is done by two additional threads that exist specifically for those purposes.

project1/thread_pool/thread_pool.h
Nickster258: > `RunAtomic` is called repeatedly until it returns a status other than `RUNNING`.

I somehow didn't catch this before, but this now makes a lot more sense to me.

PR #7, opened 2021-03-03
Add terminate messages to protobuf

Nickster258:
This adds `TerminateRequest` and `TerminateResponse`. Not to be confused by the awkward naming, `TerminateResponse` is not a response to the `terminate` command, but instead (at the moment) a response to GET and PUT requests. This response is sent before the server starts to handle the GET and PUT handling.

In working with @jake-chandler on this, we theorized that it may be worth generalizing `TerminateResponse` for most all commands sent to the server. In doing so, we could make a task utilizing the thread pool that, in its `SetUp()` phase, returns a `TerminateResponse`, so the client can handle the response accordingly. The task would then execute the command as usual, except in its `RunAtomic()`.

### DIFF
--- a/project1/wire_protocol/proto/ftp_messages.proto
+++ b/project1/wire_protocol/proto/ftp_messages.proto
@@ -10,16 +10,19 @@ message GetRequest {
 
 /// Response from the server to getting a file.
 message GetResponse {
-  /// The binary contents of the file.
-  bytes file_contents = 1;
+  /// The command ID of the associate GET request.
+  uint32 command_id = 1;
 }
 
 /// Request to the server to write a file.
 message PutRequest {
   /// The name of the file to write.
   string filename = 1;
-  /// The contents of the file to write.
-  bytes file_contents = 2;
+}
+
+message PutResponse {
+  /// The command ID of the associate PUT request.
+  uint32 command_id = 1;
 }
 
 /// Request to the server to delete a file.
@@ -63,16 +66,10 @@ message PwdResponse {
 /// Request to end the session.
 message QuitRequest {}
 
-/// Request to terminate a prior command.
-message TerminateRequest {
-  /// The command ID of the command to be terminated.
-  uint32 command_id = 1;
-}
-
-/// Response to GET and PUT with the terminate ID.
-message TerminateResponse {
-  /// The command ID of the associate GET or PUT request.
-  uint32 command_id = 1;
+/// Bidirectional message for file contents
+message FileContents {
+  /// The contents of the file
+  bytes contents = 1;
 }
 
 /**
@@ -89,7 +86,6 @@ message Request {
     MakeDirRequest make_dir = 6;
     PwdRequest pwd = 7;
     QuitRequest quit = 8;
-    TerminateRequest terminate = 9;
   }
 }
 
@@ -102,6 +98,6 @@ message Response {
     GetResponse get = 1;
     ListResponse list = 2;
     PwdResponse pwd = 3;
-    TerminateResponse terminate = 4;
+    PutResponse put = 4;
   }
 }
--- a/project1/wire_protocol/proto/ftp_messages.proto
+++ b/project1/wire_protocol/proto/ftp_messages.proto
@@ -63,6 +63,18 @@ message PwdResponse {
 /// Request to end the session.
 message QuitRequest {}
 
+/// Request to terminate a prior command.
+message TerminateRequest {
+  /// The command ID of the command to be terminated.
+  uint32 command_id = 1;
+}
+
+/// Response to GET and PUT with the terminate ID.
+message TerminateResponse {
+  /// The command ID of the associate GET or PUT request.
+  uint32 command_id = 1;
+}
+
 /**
  * @brief Captures all possible request types. This is so we can simply parse this
  * message type from the network and worry later about the exact request type.
@@ -77,6 +89,7 @@ message Request {
     MakeDirRequest make_dir = 6;
     PwdRequest pwd = 7;
     QuitRequest quit = 8;
+    TerminateRequest terminate = 9;
   }
 }
 
@@ -89,5 +102,6 @@ message Response {
     GetResponse get = 1;
     ListResponse list = 2;
     PwdResponse pwd = 3;
+    TerminateResponse terminate = 4;
   }
 }


project1/wire_protocol/proto/ftp_messages.proto
```diff
@@ -63,6 +63,18 @@ message PwdResponse {
...
 message QuitRequest {}
 
+/// Request to terminate a prior command.
+message TerminateRequest {
+  /// The command ID of the command to be terminated.
+  uint32 command_id = 1;
+}
+
+/// Response to GET and PUT with the terminate ID.
+message TerminateResponse {
```
djpetti: I'm not really a fan of this naming. Based on the way the other messages are named, `TerminateResponse` implies that it's a response for a specific `TerminateRequest`.

Also, I'm not sure that even having this as a separate message is the correct way to do it. For GET actions, that would break the one-message-per-response paradigm that we've been using so far, because you'd have a `GetReponse` from the server and then a `TerminateResponse` immediately afterward.

How about instead of having this as a separate message, you add `command_id` parameter to the `GetResponse` message? (You would also have to create a `PutResponse` message with this single parameter.)

djpetti (changes requested):

djpetti:
> In working with @jake-chandler on this, we theorized that it may be worth generalizing `TerminateResponse` for most all commands sent to the server. In doing so, we could make a task utilizing the thread pool that, in its `SetUp()` phase, returns a `TerminateResponse`, so the client can handle the response accordingly. The task would then execute the command as usual, except in its `RunAtomic()`.

Two things: 
- The thread pool is currently not really designed to support returning actual results from `SetUp`. That could be changed, but I'm not convinced that this is the correct way to go about it.
- Currently, I don't see a clear reason to make it the main thread's responsibility to actually send the `TerminateResponse` to the client. To me, it seems like doing it that way would just make Jake's life more difficult, because he'd have to split functionality between the `Agent` and `Server` classes. I would make the main thread completely agnostic to the details of the communication protocol between the client and server. Basically, the main thread spawns a sub-thread to handle each client, joins the thread when the client disconnects (or the operation is cancelled), and ignores it in-between. In other words, the `Agent` class would be solely responsible for communicating the command ID to the client.

Note that this does not preclude the main thread being _aware_ of command IDs, which might indeed be necessary to implement cancellation. You're probably going to have to implement some global thread-safe data structure to store command IDs anyway, so that the main thread can have access to them for cancellation purposes.

Or, ignoring what I said earlier, you could have one thread _per command_ instead of per client, and have the main thread choose the thread ID when it spawns off the sub-thread. Food for thought.

project1/wire_protocol/proto/ftp_messages.proto
Nickster258: > I'm not really a fan of this naming. Based on the way the other messages are named, `TerminateResponse` implies that it's a response for a specific `TerminateRequest`.

I am also very much not a fan of the naming, but wasn't sure of what else it could be called. Do you have any suggestions?

> Also, I'm not sure that even having this as a separate message is the correct way to do it. For GET actions, that would break the one-message-per-response paradigm that we've been using so far, because you'd have a `GetReponse` from the server and then a `TerminateResponse` immediately afterward.

With our idea, it would send the `TerminateResponse` first, before sending any other response. This way, the client can prepare itself for a forked `GetResponse` to handle as a task on the thread pool.

> How about instead of having this as a separate message, you add `command_id` parameter to the `GetResponse` message? (You would also have to create a `PutResponse` message with this single parameter.)

We initially planned on doing it like that, but we changed our minds based on our understanding of how protobuf works. We could be wrong, but we're under the impression that none of the response could be interpreted until all of the response is received by the client. In the case of a `GetResponse`, file contents would be received at the same time as the terminate ID, voiding the entire purpose of that terminate ID. However, we may be wrong as we haven't explicitly tested that.

djpetti: Yeah, actually, that didn't occur to me. I think you're right that we need two separate responses if we want to preempt the file data with the command ID.

However, I would be more in favor of doing it the reverse way: Create a separate `FileContents` message for holding the contents of large files, and still send the command ID in `GetResponse` and `PutResponse` messages. That entirely side-steps the naming issue, and I think it's more intuitive.

Side note: The "correct" way to deal with large files is to break them up into chunks and send each individually. If you don't do it that way, at some point you have to start considering RAM usage on the client and server side. It really depends on exactly how big the files that this is going to be tested with are. Might be a good question for the professor...

Nickster258:
> The thread pool is currently not really designed to support returning actual results from `SetUp`. That could be changed, but I'm not convinced that this is the correct way to go about it.
>
> Currently, I don't see a clear reason to make it the main thread's responsibility to actually send the `TerminateResponse` to the client. To me, it seems like doing it that way would just make Jake's life more difficult, because he'd have to split functionality between the `Agent` and `Server` classes. I would make the main thread completely agnostic to the details of the communication protocol between the client and server. Basically, the main thread spawns a sub-thread to handle each client, joins the thread when the client disconnects (or the operation is cancelled), and ignores it in-between. In other words, the `Agent` class would be solely responsible for communicating the command ID to the client.

I see now that my wording for that paragraph was rather poor. The concept here is that the server, whether it be the agent or the server itself, has a generalized `CommandExecutor` task (or something of that nature). This task (or the process that instantiates the task) would generate a `TerminateResponse` message in the `SetUp()` phase and send that to the client. It would then execte the command as it would have done previously, except within the task. The agent would then resume listening to more messages as per usual. However, if the server sees a `TerminateRequest` with an ID associated to a previous request, it could interact with the thread pool accordingly, terminating the associated task.

One of the theorized benefits of doing it like this was that it could generalize all of the functionality for the server. Whether or not the client wanted the request forked or not, it is all the same for the server.

project1/wire_protocol/proto/ftp_messages.proto
Nickster258: > However, I would be more in favor of doing it the reverse way: Create a separate `FileContents` message for holding the contents of large files, and still send the command ID in `GetResponse` and `PutResponse` messages. That entirely side-steps the naming issue, and I think it's more intuitive.

I actually really like this idea, though the `FileContents` message may have to include the command ID also to resolve potential concurrency issues.

djpetti:
> I see now that my wording for that paragraph was rather poor. The concept here is that the server, whether it be the agent or the server itself, has a generalized `CommandExecutor` task (or something of that nature). This task (or the process that instantiates the task) would generate a `TerminateResponse` message in the `SetUp()` phase and send that to the client. It would then execte the command as it would have done previously, except within the task. The agent would then resume listening to more messages as per usual. However, if the server sees a `TerminateRequest` with an ID associated to a previous request, it could interact with the thread pool accordingly, terminating the associated task.
> 
> One of the theorized benefits of doing it like this was that it could generalize all of the functionality for the server. Whether or not the client wanted the request forked or not, it is all the same for the server.

Oh, okay. That makes more sense. I'm assuming that the way this would be implemented in real life is that you would have some kind of `CancellableCommandTask` class that inherits from `Task` and implements the logic for sending the command ID before doing its actual work.

project1/wire_protocol/proto/ftp_messages.proto
djpetti: Yeah, I'd go ahead and assume that it doesn't for now and add that later if necessary.

PR #8, opened 2021-03-06
Add support for concurrent file access.

djpetti:
This also eliminates `FileHandler's` reliance on the current
directory, allowing it to safely be used with multiple concurrent
connections.

Also, I have added some tests for this functionality.

### DIFF
--- a/project1/server/file_handler/file_handler.h
+++ b/project1/server/file_handler/file_handler.h
@@ -8,6 +8,7 @@
 #include <sys/stat.h>
 #include <unistd.h>
 
+#include <filesystem>
 #include <fstream>
 #include <iostream>
 #include <vector>
@@ -21,13 +22,15 @@
 namespace server::file_handler {
 class FileHandler : public IFileHandler {
  public:
+  FileHandler();
+
   [[nodiscard]] std::vector<uint8_t> Get(
-      const std::string& filename) const final;
+      const std::string& filename) const override;
 
   bool Put(const std::string& filename,
-           const std::vector<uint8_t>& contents) final;
+           const std::vector<uint8_t>& contents) override;
 
-  bool Delete(const std::string& filename) final;
+  bool Delete(const std::string& filename) override;
 
   [[nodiscard]] std::vector<std::string> List() const final;
 
@@ -35,9 +38,13 @@ class FileHandler : public IFileHandler {
 
   bool UpDir() final;
 
-  bool MakeDir(const std::string& name) final;
+  bool MakeDir(const std::string& name) override;
 
   [[nodiscard]] std::string GetCurrentDir() const final;
+
+ private:
+  /// Keeps track of the current directory.
+  std::filesystem::path current_dir_;
 };
 }  // namespace server::file_handler
 #endif  // PROJECT1_FILE_HANDLER_H
--- a/project1/server/file_handler/tests/test_thread_safe_file_handler.cpp
+++ b/project1/server/file_handler/tests/test_thread_safe_file_handler.cpp
@@ -1,0 +1,222 @@
+#include <cstdint>
+#include <ctime>
+#include <filesystem>
+#include <memory>
+#include <random>
+#include <sstream>
+#include <thread>
+#include <vector>
+
+#include "../file_access_manager.h"
+#include "../thread_safe_file_handler.h"
+#include "gtest/gtest.h"
+
+using std::filesystem::path;
+
+namespace server::file_handler::tests {
+namespace {
+
+/**
+ * @brief Manages a temp directory to use for testing.
+ */
+class TestDir {
+ public:
+  TestDir() {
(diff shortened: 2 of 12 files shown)


project1/server/file_handler/thread_safe_file_handler.cpp
```diff
@@ -0,0 +1,52 @@
+#include "thread_safe_file_handler.h"
+
+#include <utility>
+
+#include "file_lock_guard.h"
+
+namespace server::file_handler {
+
+ThreadSafeFileHandler::ThreadSafeFileHandler(
```
jake-chandler, line 9: I don't really understand this syntax for the constructor. What is this kind of constructor definition called so I can read more documentation about it?

project1/server/file_handler/tests/test_thread_safe_file_handler.cpp
```diff
@@ -0,0 +1,222 @@
...
+  [[nodiscard]] path Get() const { return test_dir_; }
+
+ private:
+  /// Directory to use for storing the tests.
+  path test_dir_;
+};
+
+/**
+ * @brief Encapsulates the FileAccessManagers we are using for these tests.
+ */
```
jake-chandler, line 54: So, the idea here is that all threads that would perform a `file_handler` operation will share the same `AccessManagers` ?

jake-chandler (approved):
Looking through this, it seems to me that the few changes I'll have to make are minimal, but namely:

(1) Use one instance of `ThreadSafeFileHandler` per thread. In practice this will be each `Agent` having it's own instance of the `ThreadSafeFileHandler`.
(2)  Instantiation of `ThreadSafeFileHandler` requires 2 unique instances of a `FileAccessManager`, one for reading, one for writing. 
(3) All instances of `ThreadSafeFileHandler` will share the same instances of the `FileAccessManager`.

I think these are the key points I need to make note of when changing the `Agent` class. If I'm missing any other points, let me know. Otherwise, I approve.

project1/server/file_handler/thread_safe_file_handler.h
```diff
@@ -0,0 +1,53 @@
...
+   * directory.
+   * @param path The relative path to use.
+   * @return The absolute path.
+   */
+  [[nodiscard]] std::filesystem::path ToAbsolute(
+      const std::filesystem::path &path) const;
+
+  /// Synchronizes read access to files from multiple threads.
+  std::shared_ptr<FileAccessManager> read_manager_;
+  /// Synchronizes write access to files from multiple threads.
```
Nickster258, line 47: At a glance, I'd assume this would only need one `FileAccessManager` to lock and unlock appropriate files. Why is there a `read_manager_` and a `write_manager_`?

project1/server/file_handler/file_handler.h
```diff
@@ -21,23 +22,29 @@
 namespace server::file_handler {
 class FileHandler : public IFileHandler {
  public:
+  FileHandler();
+
   [[nodiscard]] std::vector<uint8_t> Get(
-      const std::string& filename) const final;
+      const std::string& filename) const override;
```
Nickster258, line 28: I am probably missing something here, but why did these go from `final` to `override`?

Nickster258 (commented):
Other than my questions for clarification, everything looks good. I appreciate the test cases as it looks relatively simple to implement.

project1/server/file_handler/thread_safe_file_handler.cpp
djpetti: What syntax specifically? The [initialization list](https://en.cppreference.com/w/cpp/language/constructor)?

project1/server/file_handler/tests/test_thread_safe_file_handler.cpp
djpetti: Yes, that is a necessity.

djpetti:
> Looking through this, it seems to me that the few changes I'll have to make are minimal, but namely:
> 
> (1) Use one instance of `ThreadSafeFileHandler` per thread. In practice this will be each `Agent` having it's own instance of the `ThreadSafeFileHandler`.
> (2) Instantiation of `ThreadSafeFileHandler` requires 2 unique instances of a `FileAccessManager`, one for reading, one for writing.
> (3) All instances of `ThreadSafeFileHandler` will share the same instances of the `FileAccessManager`.
> 
> I think these are the key points I need to make note of when changing the `Agent` class. If I'm missing any other points, let me know. Otherwise, I approve.

I believe you've got the usage right.

project1/server/file_handler/file_handler.h
djpetti: `final` prevents you from overriding that method in a subclass. `override` does not. I usually prefer `final` and only fall back on the more permissive `override` when it's actually necessary.

PR #9, opened 2021-03-07
Feature/server tasks

jake-chandler:
This is _mostly_ complete I believe. I think there could be some issues with how I'm passing down `active_command_ids` data structure (i.e. using `shared_pointer` incorrectly). 

Similarly I use a `shared_pointer` to pass the `FileAccessManagers` down. 

I'm thinking the `CommandIDs' class to house the data structure for active commands could not be completely thread safe (I'm only using c++ lock guards before data accesses).

If i've misinterpreted how the `while (parser_.hasCompleteMessage())` works there will be an issue with the checking for termination after reading 1000 bytes.

Similarly, if i've misinterpreted the use of `outgoing_message_buffer_.size()`, there will be an issue with the checking for termination after sending 1000 bytes. 

I believe I need to implement something to tell the the `agent_task` to stop running when a client disconnects.

I don't particularly like how I have to call other methods on certain `Tasks` before adding them to the pool (This breaks the point of the set up function maybe?)

Other than what I've just mentioned, I believe everything else should be sound.
Looking forward to reading your thoughts.

### DIFF
--- a/project1/server/CMakeLists.txt
+++ b/project1/server/CMakeLists.txt
@@ -1,7 +1,7 @@
-add_subdirectory(file_handler)
+add_subdirectory(file_handler server_tasks)
 
 add_library(server_lib server.cpp agent.cpp)
-target_link_libraries(server_lib wire_protocol file_handler)
+target_link_libraries(server_lib wire_protocol file_handler thread_pool)
 
-add_executable(server server_main.cpp)
+add_executable(server server_main.cpp server_tasks/nport_task.h server_tasks/nport_task.cpp server_tasks/agent_task.h server_tasks/agent_task.cpp server_tasks/tport_task.h server_tasks/tport_task.cpp server_tasks/command_ids.h server_tasks/command_ids.cpp server_tasks/server_task.h server_tasks/server_task.cpp)
 target_link_libraries(server server_lib)
--- a/project1/server/CMakeLists.txt
+++ b/project1/server/CMakeLists.txt
@@ -3,5 +3,5 @@ add_subdirectory(file_handler server_tasks)
 add_library(server_lib server.cpp agent.cpp)
 target_link_libraries(server_lib wire_protocol file_handler thread_pool)
 
-add_executable(server server_main.cpp server_tasks/nport_task.h server_tasks/nport_task.cpp server_tasks/agent_task.h server_tasks/agent_task.cpp server_tasks/tport_task.h server_tasks/tport_task.cpp server_tasks/command_ids.h server_tasks/command_ids.cpp server_tasks/server_task.h server_tasks/server_task.cpp)
+add_executable(server server_main.cpp server_tasks/nport_task.cpp server_tasks/agent_task.cpp server_tasks/tport_task.cpp server_tasks/command_ids.cpp server_tasks/server_task.cpp)
 target_link_libraries(server server_lib)
--- a/project1/server/agent.cpp
+++ b/project1/server/agent.cpp
@@ -6,6 +6,8 @@
 #include <cstdio>
 #include "stdlib.h"
 #include <iostream>
+#include <algorithm>
+#include <limits>
 
 namespace server {
 
@@ -180,15 +182,18 @@ namespace server {
         }
 
         // monitor for termination request for get and put commands
-        uint16_t kBytesSent = 0;
-        const uint32_t kBytesToSend = 1000;
+        uint16_t total_bytes_sent = 0;
         bool terminated = false;
 
         // continue until we've sent the entire message
-        while (kBytesSent < outgoing_message_buffer_.size()) {
+        while (total_bytes_sent < outgoing_message_buffer_.size()) {
             if (!terminated) {
-                kBytesSent += send(client_fd_, outgoing_message_buffer_.data() + kBytesSent,
-                                   kBytesToSend, 0);
+                const uint32_t kBytesToSend = std::min( (int) outgoing_message_buffer_.size() - total_bytes_sent, 1000);
+                int bytes_sent =  send(client_fd_, outgoing_message_buffer_.data() + total_bytes_sent,
+                                       kBytesToSend, 0);
+                if (bytes_sent != -1) {
+                    total_bytes_sent+=bytes_sent;
+                }
             } else {
                 return false;
             }
@@ -346,7 +351,7 @@ namespace server {
         int cmd_int(std::stoi(cmd_str));
         uint16_t cmd_id(0);
 
-        if (cmd_int <= static_cast<int>(UINT16_MAX) && cmd_int >= 0) {
+        if (cmd_int <= std::numeric_limits<uint16_t>::max() && cmd_int >= 0) {
             cmd_id = static_cast<uint16_t>(cmd_int);
         } else {
             std::cout << "Casting Error. " << std::endl;
--- a/project1/server/agent.cpp
+++ b/project1/server/agent.cpp
@@ -183,14 +183,15 @@ namespace server {
 
         // monitor for termination request for get and put commands
         uint32_t total_bytes_sent = 0;
+        uint32_t constexpr kChunkSize = 1000;
         bool terminated = false;
 
         // continue until we've sent the entire message
         while (total_bytes_sent < outgoing_message_buffer_.size()) {
             if (!terminated) {
-                const uint32_t kSendChunkSize = std::min((int) outgoing_message_buffer_.size() - total_bytes_sent, kSendChunkSize);
+                const uint32_t kBytesToSend = std::min((int) outgoing_message_buffer_.size() - total_bytes_sent, kChunkSize);
                 int bytes_sent =  send(client_fd_, outgoing_message_buffer_.data() + total_bytes_sent,
-                                       kSendChunkSize, 0);
+                                       kBytesToSend, 0);
                 if (bytes_sent != -1) {
                     total_bytes_sent+=bytes_sent;
                 }
--- a/project1/server/agent.cpp
+++ b/project1/server/agent.cpp
@@ -185,19 +185,19 @@ namespace server {
         }
 
         // monitor for termination request for get and put commands
-        uint16_t totalBytesSent = 0;
+        uint16_t total_bytes_sent = 0;
 
         // make 1000 byte buffer
-        int numBytesToSend = 1000;
-        uint8_t buf[numBytesToSend];
+        int num_bytes_to_send = 1000;
+        uint8_t buf[num_bytes_to_send];
 
-        // counter used to track
+        // counter used to track when buf is filled
         int byte_counter = 0;
 
         bool terminated = false;
 
         // continue until we've sent the entire message
-        while (totalBytesSent < outgoing_message_buffer_.size()) {
+        while (total_bytes_sent < outgoing_message_buffer_.size()) {
 
 
             // fill the buffer one byte at a time
@@ -211,7 +211,7 @@ namespace server {
                     if (!terminated) {
                         // Send 1000 bytes of the message.
                         if (send(client_fd_, buf,
-                                 numBytesToSend, 0) < 0) {
+                                 num_bytes_to_send, 0) < 0) {
                             perror("Failed to send the file contents.");
                             return false;
                         }
@@ -223,10 +223,14 @@ namespace server {
                     byte_counter = 0;
 
                     // update
-                    totalBytesSent += 1000;
+                    total_bytes_sent += 1000;
                 }
                 buf[byte_counter] = outgoing_message_buffer_.at(n);
                 ++byte_counter;
+
+                if (active_commands_->Contains(command_id)){
+
+                }
 
             }
         }
--- a/project1/server/agent.cpp
+++ b/project1/server/agent.cpp
@@ -20,20 +20,23 @@ namespace server {
 
     }  // namespace
 
-    Agent::Agent(int client_fd, std::unique_ptr<ThreadSafeFileHandler> file_handler)
-            : client_fd_(client_fd), file_handler_(std::move(file_handler)) {}
+    Agent::Agent(int client_fd, std::unique_ptr<ThreadSafeFileHandler> file_handler,
+                 std::shared_ptr<server_tasks::CommandIDs> active_commands)
+            : client_fd_(client_fd), active_commands_(active_commands), file_handler_(std::move(file_handler)) {}
+
+    Agent::Agent(int client_fd,
+                 std::shared_ptr<server_tasks::CommandIDs> active_commands) :
+            client_fd_(client_fd), active_commands_(active_commands) {}
 
     Agent::~Agent() {
         // Close the socket.
         close(client_fd_);
     }
 
-    void Agent::SetActiveCommandIDs(std::shared_ptr<server_tasks::CommandIDs> cmd_ids) {
-        active_commands_ = cmd_ids;
-    }
-
     bool Agent::Handle() {
         ClientState client_state = ClientState::ACTIVE;
+
+
         while (client_state == ClientState::ACTIVE) {
             // Get the next message from the socket.
             Request message;
@@ -48,17 +51,16 @@ namespace server {
         return client_state != ClientState::ERROR;
     }
 
-    Agent::ClientState Agent::ReadFileContents(ftp_messages::FileContents *fc) {
+    Agent::ClientState Agent::ReadFileContents(ftp_messages::FileContents *fc, uint16_t command_id) {
 
-        bool terminated = false;
-        while (!fc_parser_.HasCompleteMessage()){
+        while (!fc_parser_.HasCompleteMessage()) {
             incoming_message_buffer_.resize(kClientBufferSize);
 
             uint8_t buf[1000];
 
             // Read 1000 bytes from the socket
             const auto bytes_read =
-                    recv(client_fd_, buf ,1000, 0);
+                    recv(client_fd_, buf, 1000, 0);
 
             if (bytes_read < 0) {
                 // Failed to read anything.
@@ -71,6 +73,8 @@ namespace server {
                 return ClientState::DISCONNECTED;
             }
 
+            bool terminated = !active_commands_->Contains(command_id);
+
             // only continue if this put request is an active command
             if (!terminated) {
 
@@ -79,6 +83,8 @@ namespace server {
                     incoming_message_buffer_.push_back(buf[i]);
                 }
 
+            } else {
+                std::cout << "Command #" << command_id << " successfully terminated." << std::endl;
             }
 
             // The parser assumes that the entire vector contains valid data, so limit
@@ -95,6 +101,7 @@ namespace server {
         }
         return ClientState::ACTIVE;
     }
+
     Agent::ClientState Agent::ReadNextMessage(Request *message) {
         while (!parser_.HasCompleteMessage()) {
             incoming_message_buffer_.resize(kClientBufferSize);
@@ -146,6 +153,8 @@ namespace server {
             return HandleRequest(message.pwd());
         } else if (message.has_quit()) {
             return HandleRequest(message.quit());
+        } else if (message.has_terminate()) {
+            return HandleRequest(message.terminate());
         }
 
         std::cerr << "No valid message was received." << std::endl;
@@ -167,7 +176,8 @@ namespace server {
         }
         return true;
     }
-    bool Agent::SendFileContents(const ftp_messages::FileContents &file_contents) {
+
+    bool Agent::SendFileContents(const ftp_messages::FileContents &file_contents, uint16_t command_id) {
         // Serialize the message.
         if (!wire_protocol::Serialize(file_contents, &outgoing_message_buffer_)) {
             std::cerr << "Failed to serialize message." << std::endl;
@@ -184,6 +194,8 @@ namespace server {
         // counter used to track
         int byte_counter = 0;
 
+        bool terminated = false;
+
         // continue until we've sent the entire message
         while (totalBytesSent < outgoing_message_buffer_.size()) {
 
@@ -193,23 +205,24 @@ namespace server {
 
                 // check if 1000 byte buffer is filled
                 if (byte_counter == 1000) {
-                    bool terminated = false;
+                    terminated = !active_commands_->Contains(command_id);
 
                     // check if the process has been terminated
                     if (!terminated) {
                         // Send 1000 bytes of the message.
                         if (send(client_fd_, buf,
                                  numBytesToSend, 0) < 0) {
-                            perror("Failed to send the message");
+                            perror("Failed to send the file contents.");
                             return false;
                         }
-                    }
-                    else {
-
+                    } else {
+                        return false;
                     }
 
                     // reset
                     byte_counter = 0;
+
+                    // update
                     totalBytesSent += 1000;
                 }
                 buf[byte_counter] = outgoing_message_buffer_.at(n);
@@ -230,7 +243,7 @@ namespace server {
         uint32_t id;
         do {
             // ensure this id is not already in use
-            id = rand()%1000000 + 1;
+            id = rand() % 1000000 + 1;
         } while (active_commands_->Contains(id));
 
         // register this command as an active command
@@ -252,7 +265,7 @@ namespace server {
 
         fc.set_contents(file_data.data(), file_data.size());
 
-        return SendFileContents(fc) ? ClientState::ACTIVE : ClientState::ERROR;
+        return SendFileContents(fc, id) ? ClientState::ACTIVE : ClientState::ERROR;
     }
 
     Agent::ClientState Agent::HandleRequest(
@@ -262,7 +275,7 @@ namespace server {
         Response r;
         uint32_t id;
         do {
-            id = rand()%1000000 + 1;
+            id = rand() % 1000000 + 1;
         } while (active_commands_->Contains(id));
 
         r.mutable_put()->set_command_id(id);
@@ -274,7 +287,7 @@ namespace server {
         ftp_messages::FileContents fc;
 
         //Retrieve the file contents
-        ReadFileContents(&fc);
+        ReadFileContents(&fc, id);
 
         // Write the file data.
         std::vector<uint8_t> file_contents(fc.contents().begin(),
@@ -369,10 +382,37 @@ namespace server {
     }
 
     Agent::ClientState Agent::HandleRequest(
+            const ftp_messages::TerminateRequest &request) {
+        std::cout<< "Terminating an active GET or PUT command." << std::endl;
+
+        // Retrieve the command id.
+        std::string cmd_str = request.command_id();
+
+        // Case to uint16_t
+        int cmd_int(std::stoi(cmd_str));
+        uint16_t cmd_id(0);
+
+        if (cmd_int <= static_cast<int>(UINT16_MAX) && cmd_int >=0) {
+            cmd_id = static_cast<uint16_t>(cmd_int);
+        }
+        else {
+            std::cout << "Casting Error. " << std::endl;
+        }
+
+        // remove command from active command list.
+        active_commands_->Delete(cmd_id);
+
+        return ClientState::ACTIVE;
+
+    }
+
+    Agent::ClientState Agent::HandleRequest(
             const ftp_messages::QuitRequest &request) {
         std::cout << "Got a QUIT request. Exiting." << std::endl;
         // Indicate that we are finished with this client.
         return ClientState::DISCONNECTED;
     }
+
+
 
 }  // namespace server
--- a/project1/server/agent.cpp
+++ b/project1/server/agent.cpp
@@ -4,231 +4,375 @@
 #include <unistd.h>
 
 #include <cstdio>
+#include "stdlib.h"
 #include <iostream>
 
 namespace server {
 
-using file_handler::IFileHandler;
-using ftp_messages::Request;
-using ftp_messages::Response;
+    using file_handler::ThreadSafeFileHandler;
+    using ftp_messages::Request;
+    using ftp_messages::Response;
 
-namespace {
+    namespace {
 
-/// Generic empty response.
-const Response kEmptyResponse{};
+        /// Generic empty response.
+        const Response kEmptyResponse{};
 
-}  // namespace
+    }  // namespace
 
-Agent::Agent(int client_fd, std::unique_ptr<IFileHandler> file_handler)
-    : client_fd_(client_fd), file_handler_(std::move(file_handler)) {}
+    Agent::Agent(int client_fd, std::unique_ptr<ThreadSafeFileHandler> file_handler)
+            : client_fd_(client_fd), file_handler_(std::move(file_handler)) {}
 
-Agent::~Agent() {
-  // Close the socket.
-  close(client_fd_);
-}
-
-bool Agent::Handle() {
-  ClientState client_state = ClientState::ACTIVE;
-  while (client_state == ClientState::ACTIVE) {
-    // Get the next message from the socket.
-    Request message;
-    client_state = ReadNextMessage(&message);
-
-    if (client_state == ClientState::ACTIVE) {
-      // Handle the message.
-      client_state = DispatchMessage(message);
-    }
-  }
-
-  return client_state != ClientState::ERROR;
-}
-
-Agent::ClientState Agent::ReadNextMessage(Request* message) {
-  while (!parser_.HasCompleteMessage()) {
-    incoming_message_buffer_.resize(kClientBufferSize);
-
-    // Read some more data from the socket.
-    const auto bytes_read =
-        recv(client_fd_, incoming_message_buffer_.data(), kClientBufferSize, 0);
-    if (bytes_read < 0) {
-      // Failed to read anything.
-      perror("Failed to read from client socket");
-      parser_.ResetParser();
-      return ClientState::ERROR;
-    } else if (bytes_read == 0) {
-      // Client has disconnected nicely.
-      std::cout << "Detected client disconnect." << std::endl;
-      return ClientState::DISCONNECTED;
+    Agent::~Agent() {
+        // Close the socket.
+        close(client_fd_);
     }
 
-    // The parser assumes that the entire vector contains valid data, so limit
-    // the size.
-    incoming_message_buffer_.resize(bytes_read);
-
-    // Add the data to the parser.
-    parser_.AddNewData(incoming_message_buffer_);
-  }
-
-  // Get the parsed message.
-  if (!parser_.GetMessage(message)) {
-    std::cerr << "Failed to get the parsed message." << std::endl;
-    return ClientState::ERROR;
-  }
-  return ClientState::ACTIVE;
-}
-
-Agent::ClientState Agent::DispatchMessage(const Request& message) {
-  if (message.has_get()) {
-    return HandleRequest(message.get());
-  } else if (message.has_put()) {
-    return HandleRequest(message.put());
-  } else if (message.has_delete_()) {
-    return HandleRequest(message.delete_());
-  } else if (message.has_list()) {
-    return HandleRequest(message.list());
-  } else if (message.has_change_dir()) {
-    return HandleRequest(message.change_dir());
-  } else if (message.has_make_dir()) {
-    return HandleRequest(message.make_dir());
-  } else if (message.has_pwd()) {
-    return HandleRequest(message.pwd());
-  } else if (message.has_quit()) {
-    return HandleRequest(message.quit());
-  }
-
-  std::cerr << "No valid message was received." << std::endl;
-  return ClientState::ERROR;
-}
-
-bool Agent::SendResponse(const Response& response) {
-  // Serialize the message.
-  if (!wire_protocol::Serialize(response, &outgoing_message_buffer_)) {
-    std::cerr << "Failed to serialize message." << std::endl;
-    return false;
-  }
-
-  // Send the message.
-  if (send(client_fd_, outgoing_message_buffer_.data(),
-           outgoing_message_buffer_.size(), 0) < 0) {
-    perror("Failed to send the message");
-    return false;
-  }
-  return true;
-}
-
-Agent::ClientState Agent::HandleRequest(
-    const ftp_messages::GetRequest& request) {
-  std::cout << "Performing a GET operation." << std::endl;
-
-  // Get the file data.
-  const auto file_data = file_handler_->Get(request.filename());
-
-  // Send the response.
-  Response response;
-  response.mutable_get()->set_file_contents(file_data.data(), file_data.size());
-  return SendResponse(response) ? ClientState::ACTIVE : ClientState::ERROR;
-}
-
-Agent::ClientState Agent::HandleRequest(
-    const ftp_messages::PutRequest& request) {
-  std::cout << "Performing a PUT operation." << std::endl;
-
-  // Write the file data.
-  std::vector<uint8_t> file_contents(request.file_contents().begin(),
-                                     request.file_contents().end());
-  if (!file_handler_->Put(request.filename(), file_contents)) {
-    std::cerr << "Failed to write the file." << std::endl;
-    return ClientState::ERROR;
-  }
-
-  return SendResponse(kEmptyResponse) ? ClientState::ACTIVE
-                                      : ClientState::ERROR;
-}
-
-Agent::ClientState Agent::HandleRequest(
-    const ftp_messages::DeleteRequest& request) {
-  std::cout << "Performing a DELETE request." << std::endl;
-
-  // Delete the file.
-  if (!file_handler_->Delete(request.filename())) {
-    std::cerr << "Failed to delete the file." << std::endl;
-    return ClientState::ERROR;
-  }
-
-  return SendResponse(kEmptyResponse) ? ClientState::ACTIVE
-                                      : ClientState::ERROR;
-}
-
-Agent::ClientState Agent::HandleRequest(
-    const ftp_messages::ListRequest& request) {
-  std::cout << "Performing a LIST request." << std::endl;
-
-  // List the directory contents.
-  const auto& kDirectoryContents = file_handler_->List();
-
-  // Send the response.
-  Response response;
-  response.mutable_list()->clear_filenames();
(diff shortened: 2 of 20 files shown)


project1/server/CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
-add_subdirectory(file_handler)
+add_subdirectory(file_handler server_tasks)
 
 add_library(server_lib server.cpp agent.cpp)
-target_link_libraries(server_lib wire_protocol file_handler)
+target_link_libraries(server_lib wire_protocol file_handler thread_pool)
 
-add_executable(server server_main.cpp)
+add_executable(server server_main.cpp server_tasks/nport_task.h server_tasks/nport_task.cpp server_tasks/agent_task.h server_tasks/agent_task.cpp server_tasks/tport_task.h server_tasks/tport_task.cpp server_tasks/command_ids.h server_tasks/command_ids.cpp server_tasks/server_task.h server_tasks/server_task.cpp)
```
djpetti: There's really no need to include `.h` files directly here. (Just told this to Nick also.)

project1/server/agent.h
```diff
@@ -12,95 +12,142 @@
...
+         * @param file_handler The `FileHandler` to use internally. Note this class
+         *    will take ownership of it.
+         */
+        Agent(int client_fd,
+              std::unique_ptr<file_handler::ThreadSafeFileHandler> file_handler,
+              std::shared_ptr<server_tasks::CommandIDs> active_commands);
+
+        /**
+         * @param client_fd The FD of the client socket. Note that it will take
+         *    responsibility for closing this socket on exit.
```
djpetti, line 41: Missing docs for `active_commands` parameter. Same issue with the other constructor.

project1/server/server.cpp
```diff
@@ -10,105 +10,32 @@
...
-  address.sin_addr.s_addr = INADDR_ANY;
-  address.sin_port = htons(port);
+        std::shared_ptr<server_tasks::CommandIDs> active_ids;
+        auto nPortTask = std::make_shared<server_tasks::NPortTask>();
+        auto tPortTask = std::make_shared<server_tasks::TPortTask>();
 
-  return address;
-}
+        // pass active command list to nPortTask and tPortTask
+        nPortTask->SetActiveCommandIDs(active_ids);
```
djpetti: Could these not just be constructor parameters?

project1/server/server.cpp
```diff
@@ -10,105 +10,32 @@
...
-  if (bind(server_fd, (struct sockaddr *)&address, sizeof(address)) < 0) {
-    perror("bind() failed on server socket");
-    return -1;
-  }
-  if (listen(server_fd, kMaxQueueSize) < 0) {
-    perror("listen() failed on server socket");
-    return -1;
-  }
+        pool.AddTask(nPortTask);
+        pool.AddTask(tPortTask);
```
djpetti, line 23: So, as you have it now, when control gets to here, it's going to destroy the pool and the tasks you just added will be immediately cancelled. You need some way of keeping the main thread alive indefinitely. I suggest you sit here and wait for all the tasks in the pool to exit naturally.

project1/server/server_main.cpp
```diff
@@ -13,22 +13,27 @@
  */
 void PrintUsageAndExit(const char* program_name) {
```
djpetti, line 14: (minor) It would be nice to initialize logging here according to [this page](https://github.com/emilk/loguru)...

project1/server/server_tasks/CMakeLists.txt
```diff
@@ -0,0 +1,3 @@
+add_library(server_tasks_lib)
```
djpetti: Ummm, don't you actually have to add at least one `.cpp` file to this library?

project1/server/server_tasks/agent_task.cpp
```diff
@@ -0,0 +1,35 @@
...
+    void AgentTask::SetFileAccessManagers(std::shared_ptr<server::file_handler::FileAccessManager> read_mgr,
+                                          std::shared_ptr<server::file_handler::FileAccessManager> write_mgr) {
+        read_manager_ = read_mgr;
+        write_manager_ = write_mgr;
+    }
+
+    thread_pool::Task::Status AgentTask::SetUp() {
+        // give the agent a unique file handler with the shared access managers
+        auto fh = std::make_unique<server::file_handler::ThreadSafeFileHandler>(read_manager_,write_manager_);
+        agent_ = new server::Agent(client_fd_, std::move(fh),active_commands_);
```
djpetti: Don't use `new`, it will lead to memory leaks as it does in this case since you never call `delete`. Use a `unique_ptr` instead.

project1/server/server_tasks/agent_task.h
```diff
@@ -0,0 +1,56 @@
...
+        thread_pool::Task::Status RunAtomic() final;
+        thread_pool::Task::Status SetUp();
+        void CleanUp() final;
+
+        /**
+         * @brief Sets the ClientFD for this agent.
+         * @param id
+         * @note Needs to be set before being added to the thread pool.
+         */
+        void SetClientFD(int id);
```
djpetti: Again, can this not just be passed in the constructor?

project1/server/server_tasks/agent_task.h
```diff
@@ -0,0 +1,56 @@
...
+
+        /**
+         * @brief Sets the list of active commands for this agent.
+         * @param cmds The list of active commands
+         * @note Needs to be set before being added to the thread pool.
+         */
+        void SetActiveCommands(std::shared_ptr<CommandIDs> cmds);
+
+        void SetFileAccessManagers(std::shared_ptr<server::file_handler::FileAccessManager> read_mgr,
```
djpetti: Missing docs

project1/server/server_tasks/command_ids.h
```diff
@@ -0,0 +1,24 @@
...
+ */
+
+#ifndef PROJECT1_COMMAND_IDS_H
+#define PROJECT1_COMMAND_IDS_H
+
+#include <mutex>
+#include <algorithm>
+#include <list>
+namespace server_tasks {
+    class CommandIDs {
```
djpetti, line 16: This whole class is missing docs (and the formatting is weird).

project1/server/server_tasks/command_ids.h
```diff
@@ -0,0 +1,24 @@
...
+    public:
+
+        void Insert(uint16_t id);
+        bool Contains(uint16_t id);
+        void Delete(uint16_t id);
+
+    private:
+        std::list<uint16_t> command_ids_;
```
djpetti: How about using an `std::unordered_set` here internally? It's more efficient, and will be easier for people reading your code to interpret.

project1/server/server_tasks/nport_task.cpp
```diff
@@ -0,0 +1,51 @@
...
+#include <sys/socket.h>
+#include <cstdio>
+#include <iostream>
+
+namespace server_tasks {
+
+    thread_pool::Task::Status NPortTask::Listen() {
+
+
+        std::cout << "Server is listening for normal commands on port " << port_ << "." << std::endl;
```
djpetti, line 14: Let's make this a `LOG_S` call, while we're changing things anyway.

project1/server/server_tasks/nport_task.cpp
```diff
@@ -0,0 +1,51 @@
...
+        while (true) {
+
+            // Accept a new connection.
+            int client_fd = accept(server_fd_, nullptr, nullptr);
+            if (client_fd < 0) {
+                perror("accept() failed");
+                return thread_pool::Task::Status::FAILED;
+            }
+
+            std::cout << "Handling new connection from client #" << client_fd << "." << std::endl;
```
djpetti, line 25: Can also be a logging call.

project1/server/server_tasks/nport_task.h
```diff
@@ -0,0 +1,40 @@
...
+#include "server_task.h"
+#include "../../thread_pool/task.h"
+#include "../../thread_pool/thread_pool.h"
+#include "command_ids.h"
+#include "../file_handler/file_handler.h"
+#include "../file_handler/file_access_manager.h"
+
+namespace server_tasks {
+
+    class NPortTask : public ServerTask {
```
djpetti, line 21: Once again, missing docs and weird formatting.

project1/server/server_tasks/server_task.cpp
```diff
@@ -0,0 +1,65 @@
...
+namespace server_tasks {
+
+    thread_pool::Task::Status ServerTask::SetUp()  {
+        // Create the socket.
+        const auto kAddress = MakeAddress(port_);
+
+        server_fd_ = SetUpSocket(kAddress);
+        if (server_fd_ < 0) {
+            return thread_pool::Task::Status::FAILED;
+        } else return thread_pool::Task::Status::RUNNING;
```
djpetti: (minor) Let's put brackets around this `else` clause. Or, you can do this with a ternary operator, e.g.
`return server_fd_ < 0 ? thread_pool::Task::Status::FAILED : thread_pool::Task::Status::RUNNING;`

project1/server/server_tasks/server_task.cpp
```diff
@@ -0,0 +1,65 @@
...
+        address.sin_port = htons(port);
+
+        return address;
+    }
+
+    int ServerTask::SetUpSocket(const struct sockaddr_in &address) {
+        // Open a TCP socket.
+        const int server_fd = socket(AF_INET, SOCK_STREAM, 0);
+        if (server_fd == 0) {
+            perror("Failed to create server socket");
```
djpetti, line 34: I know you copy/pasted this from my code, but let's still make these log messages instead of `perror`s.

project1/server/server_tasks/nport_task.h
```diff
@@ -0,0 +1,40 @@
...
+    public:
+        thread_pool::Task::Status RunAtomic() final override;
```
djpetti: Don't use both `final` and `override`. Use `final` by default, and `override` if you need to.

project1/server/server_tasks/server_task.h
```diff
@@ -0,0 +1,77 @@
...
+#include "command_ids.h"
+#include <sys/socket.h>
+#include <cstdio>
+#include <iostream>
+#include <netinet/in.h>
+#include <memory>
+#include <cstdint>
+
+namespace server_tasks {
+    class ServerTask : public thread_pool::Task {
```
djpetti, line 23: Missing class docs.

project1/server/server_tasks/server_task.h
```diff
@@ -0,0 +1,77 @@
...
+    public:
+
+        /**
+         * @brief Represents how this ServerTask will listen on a socket.
```
djpetti: (minor) Kind of a weird wording? How about simply "Listens on a socket"?

project1/server/server_tasks/server_task.h
```diff
@@ -0,0 +1,77 @@
...
+         * @return The status of the thread pool.
+         * @note This will be called in RunAtomic.
+         */
+        virtual thread_pool::Task::Status Listen();
```
djpetti: This method should probably be `protected`, not `public`.

project1/server/server_tasks/server_task.h
```diff
@@ -0,0 +1,77 @@
...
+        thread_pool::ThreadPool pool_;
+
+        ///The port # for terminate commands.
+        uint16_t port_;
+
+        ///The file descriptor for the server
+        int server_fd_;
+
+        /// Maximum queue size to use for listening on the server socket.
+        uint8_t kMaxQueueSize_= 5;
```
djpetti: (minor) I would make this `constexpr`.

project1/server/server_tasks/tport_task.cpp
```diff
@@ -0,0 +1,41 @@
...
+#include "agent_task.h"
+
+#include <sys/socket.h>
+#include <cstdio>
+#include <netinet/in.h>
+#include <cstdint>
+namespace server_tasks {
+
+    thread_pool::Task::Status TPortTask::Listen() {
+        std::cout << "Server is listening for termination commands on port " << port_ << "." << std::endl;
```
djpetti, line 12: Again, these `cout`s can be log messages.

project1/server/server_tasks/tport_task.cpp
```diff
@@ -0,0 +1,41 @@
...
+            // Accept a new connection.
+            int client_fd = accept(server_fd_, nullptr, nullptr);
+            if (client_fd < 0) {
+                perror("accept() failed");
+                break;
+            }
+
+            std::cout << "Handling new connection from client #" << client_fd << "." << std::endl;
+
+            auto agent_task = std::make_shared<AgentTask>();
```
djpetti: Wait, I thought the `TPortTask` was supposed to be terminating existing `Agents`, not making new ones? I'm confused...

project1/server/server_tasks/tport_task.h
```diff
@@ -0,0 +1,27 @@
...
+#define PROJECT1_TPORT_TASK_H
+
+#include "server_task.h"
+#include "../../thread_pool/task.h"
+#include "../../thread_pool/thread_pool.h"
+#include "command_ids.h"
+
+namespace server_tasks {
+
+    class TPortTask : public ServerTask {
```
djpetti, line 17: (minor) I know the professor calls them `nport` and `tport`, but in general, these are stupid names. Since we can name things however we want in our code, I would call these `NormalPortTask` and `TerminatePortTask`, or something along those lines.

project1/wire_protocol/proto/CMakeLists.txt
```diff
@@ -8,3 +8,6 @@ add_library(proto ${PROTO_HEADER} ${PROTO_SRC})
 target_link_libraries(proto INTERFACE ${Protobuf_LIBRARIES})
 # Make sure everything that depends on this can find the headers.
 target_include_directories(proto INTERFACE ${CMAKE_CURRENT_BINARY_DIR})
+
+# For Mac Users
+target_include_directories(proto INTERFACE "/usr/local/Cellar/protobuf/3.14.0/include")
```
djpetti: I thought we were going to wrap this in a conditional that [only ran on MacOS](https://cmake.org/cmake/help/v3.0/variable/APPLE.html)?

project1/server/agent.cpp
```diff
@@ -4,231 +4,415 @@
...
+
+        return client_state != ClientState::ERROR;
+    }
+
+    Agent::ClientState Agent::ReadFileContents(ftp_messages::FileContents *fc, uint16_t command_id) {
+
+        while (!fc_parser_.HasCompleteMessage()) {
+            incoming_message_buffer_.resize(kClientBufferSize);
+
+            uint8_t buf[1000];
```
djpetti: You might not want to allocate such a large buffer on the stack. Note that my `incoming_mesage_buffer_` variable is a vector for a reason, since vectors allocate on the heap internally. It is also a class member so that we can re-use the same buffer, as long as it's sufficiently large (hence the call to `resize` above).

Also, I don't see much of a point in having two buffers, considering that `incoming_message_buffer_` was supposed to get overwritten each time around the loop anyway. (In the new version, you don't do that, which will completely break the message parsing.) Why not just continue reading directly into there?

To summarize, I think that the only changes necessary to this method are changing `kClientBufferSize` and adding the logic to check whether we should terminate every time around the loop.

project1/server/agent.cpp
```diff
@@ -4,231 +4,415 @@
...
+
+            // Read 1000 bytes from the socket
+            const auto bytes_read =
+                    recv(client_fd_, buf, 1000, 0);
+
+            if (bytes_read < 0) {
+                // Failed to read anything.
+                perror("Failed to read from client socket");
```
djpetti, line 68: Convert to logging calls.

project1/server/agent.cpp
```diff
@@ -4,231 +4,415 @@
...
+                return ClientState::DISCONNECTED;
+            }
+
+            bool terminated = !active_commands_->Contains(command_id);
+
+            // only continue if this put request is an active command
+            if (!terminated) {
+
+                // place data into the incoming message buffer
+                for (int i = 0; i < 1000; i++) {
```
djpetti: This operation can be done more efficiently (and succinctly) with `std::copy`.

project1/server/agent.cpp
```diff
@@ -4,231 +4,415 @@
...
+                    incoming_message_buffer_.push_back(buf[i]);
+                }
+
+            } else {
+                std::cout << "Command #" << command_id << " successfully terminated." << std::endl;
```
djpetti, line 88: Hmm. In this case, you probably just want to break out of the loop. Otherwise, more likely than not, `HasCompleteMessage` will continue to return false, and the termination won't actually happen.

project1/server/agent.cpp
```diff
@@ -4,231 +4,415 @@
...
+        }
+        // Get the parsed message.
+        if (!fc_parser_.GetMessage(fc)) {
+            std::cerr << "Failed to get the parsed message." << std::endl;
+            return ClientState::ERROR;
+        }
+        return ClientState::ACTIVE;
+    }
+
+    Agent::ClientState Agent::ReadNextMessage(Request *message) {
```
djpetti, line 101: This method seems redundant, (and will only get even more so if you take my suggestions above). Would it be possible to simply combine them into one and make the termination check optional? Or, you can break out the code in the loop into a new function, and call it from both places.

project1/server/agent.cpp
```diff
@@ -4,231 +4,415 @@
...
+            std::cerr << "Failed to serialize message." << std::endl;
+            return false;
+        }
+
+        // monitor for termination request for get and put commands
+        uint16_t totalBytesSent = 0;
+
+        // make 1000 byte buffer
+        int numBytesToSend = 1000;
+        uint8_t buf[numBytesToSend];
```
djpetti: See my previous comment about not creating these on the stack.

project1/server/agent.cpp
```diff
@@ -4,231 +4,415 @@
...
+
+        // counter used to track
```
djpetti: To track what?

project1/server/agent.cpp
```diff
@@ -4,231 +4,415 @@
...
+        int byte_counter = 0;
+
+        bool terminated = false;
+
+        // continue until we've sent the entire message
+        while (totalBytesSent < outgoing_message_buffer_.size()) {
+
+
+            // fill the buffer one byte at a time
+            for (int n:outgoing_message_buffer_) {
```
djpetti: This logic, with all the copying, is kind of clunky. Could you just call `send` repeatedly with the buffer set to `outgoing_message_buffer_.data() + totalBytesSent` and `numBytesToSend`?

project1/server/agent.cpp
```diff
@@ -4,231 +4,415 @@
...
+
+    bool Agent::SendFileContents(const ftp_messages::FileContents &file_contents, uint16_t command_id) {
+        // Serialize the message.
+        if (!wire_protocol::Serialize(file_contents, &outgoing_message_buffer_)) {
+            std::cerr << "Failed to serialize message." << std::endl;
+            return false;
+        }
+
+        // monitor for termination request for get and put commands
+        uint16_t totalBytesSent = 0;
```
djpetti: There are a bunch of variables here that do not follow the style guide naming rules.

project1/server/agent.cpp
```diff
@@ -4,231 +4,415 @@
...
+        r.mutable_get()->set_command_id(id);
+
+        // Send client the command id for possible termination
+        SendResponse(r);
+
+        // Get the file data.
+        const auto file_data = file_handler_->Get(request.filename());
+
+        // Send the response.
+        Response response;
```
djpetti, line 228: (minor) Why so much vertical whitespace? Usually, I don't have a problem using vertical whitespace to group code into logical blocks, but here I'd argue that it's going too far and making things less readable.

project1/server/agent.cpp
```diff
@@ -4,231 +4,415 @@
...
+    }
+
+    Agent::ClientState Agent::HandleRequest(
+            const ftp_messages::GetRequest &request) {
+        std::cout << "Performing a GET operation." << std::endl;
+
+        Response r;
+
+        // generate a random # for the command id
+        uint32_t id;
```
djpetti: Maybe we should break this ID generation into a separate function?

djpetti (changes requested):
Quick review here... I will probably give it a second look once you fix these initial comments.

project1/server/CMakeLists.txt
jake-chandler: Gotcha. I think my CLion is set to do this by default every time a new `.h` or `.cpp` file is created...

project1/server/server.cpp
jake-chandler: How would I go about waiting all tasks to exit naturally? I see some of the examples in your tests but I'm not sure how that would apply here

project1/server/agent.cpp
jake-chandler: agreed.

project1/server/agent.cpp
jake-chandler: I'm not exactly sure what you mean by this. I figured there was a better way to go about this but I'm still not super clear on how to do that from that previous comment of yours.

project1/server/agent.cpp
jake-chandler: These will be changed.

project1/server/agent.cpp
jake-chandler: It was more appealing to my eyes when working. The vertical whitespace helps when I've been staring at the screen for awhile. Regardless, I'll remove it.

project1/server/agent.cpp
jake-chandler: I didn't mean to leave this incomplete `// counter used to track when buf is filled`

project1/server/server.cpp
jake-chandler: The only reason I didn't do that initially is because I hadn't seen a task with a constructor. For some reason this gave me the notion to avoid constructors with tasks, but I see now that's unwarrented.

project1/server/server_tasks/CMakeLists.txt
jake-chandler: Should I add all the `.cpp` files from the server_tasks? Do I even need to make this a library?

project1/server/server_tasks/agent_task.cpp
jake-chandler: So, I tried to do this initially. I kept getting a weird error with type conversion that I'm not sure how to deal with. 

This line: `agent_ = std::unique_ptr<server::Agent>(client_fd_, std::move(fh), active_commands_);`

produces this error `Assigning to 'server::Agent *' from incompatible type 'std::unique_ptr<server::Agent>'`

project1/server/server_tasks/agent_task.h
jake-chandler: I'll change these

project1/server/server_tasks/command_ids.h
jake-chandler: the `CMD-ALT-L` keybinding for formatting does this automatically. I've tried to look into changing this but I wasn't able to get it to not indent after namespace blocks...

project1/server/server.cpp
djpetti: Yep, constructors in tasks are fine. (I'm surprised you hadn't come across one before, because I definitely use them in the `ThreadPool` tests.

project1/server/server.cpp
djpetti: The trivial way is just checking the statuses of the two tasks you added in a loop every second or so and exiting when they both flip to something other than `RUNNING`. I'm also working on a better way to do this.

project1/server/agent.cpp
djpetti: Something like:
```c++
while (total_bytes_sent < outgoing_message_buffer_,size()) {
  const uint32_t kBytesToSend = std::min(outgoing_message_buffer.size() - total_bytes_sent, 1000);
  total_bytes_sent += send(client_fd_, outgoing_message_buffer.data() + total_bytes_sent, kBytesToSend);
}
```

Obviously, you would need to add error checking and all the other logic that goes on in this loop, but that's the basic idea.

project1/server/agent.cpp
djpetti: To be clear, no need to remove it _completely_. My suggestion was more along the lines of using it to group code blocks in a more logical manner.

project1/server/server_tasks/tport_task.cpp
jake-chandler: I've modified the `Agent` to handle `Termination Requests`. In a `TPortTask`, the `Agent` should only be handling `Termination Requests`. Inversely, the `NPortTask` will be handling all other requests.

project1/server/server_tasks/tport_task.h
jake-chandler: I agree. I've thought about this, and figured I'd change them after funcionality has been verified.

project1/wire_protocol/proto/CMakeLists.txt
jake-chandler: I did. I must have accidentally stashed the changes I made. Weird. I'll fix it

project1/server/agent.cpp
jake-chandler: I don't understand why you would want `kBytesToSend`set as the _minimum_ of `outgoing_message_buffer.size() - total_bytes_sent` and 1000? According to the project description, server sends/recieves 1000 bytes at a time. I'm also having trouble understanding what `outgoing_message_buffer.size() - total_bytes_sent` is supposed to represent?

project1/server/server_tasks/command_ids.cpp
```diff
@@ -0,0 +1,21 @@
...
+        command_ids_.remove(id);
+    }
+    bool CommandIDs::Contains(uint16_t id) {
+        std::lock_guard<std::mutex> guard(mutex_);
+        std::list<uint16_t>::iterator find = std::find(command_ids_.begin(),
+                                                       command_ids_.end(),id);
+        if (find != command_ids_.end()){
+            return true;
+        }
+        return false;
```
Nickster258: I don't know of our opinions on readability, but this can be
```c++
return find != command_ids_.end();
```

project1/server/server_tasks/nport_task.h
```diff
@@ -0,0 +1,45 @@
...
+                                   std::shared_ptr<server::file_handler::FileAccessManager> write_mgr);
+
+    private:
+
+        ///The file access managers. @Note Inherited from the server.
+        std::shared_ptr<server::file_handler::FileAccessManager> read_manager_;
+        std::shared_ptr<server::file_handler::FileAccessManager> write_manager_;
+
+
+
```
Nickster258, line 51: (minor) A couple of extra lines here.

project1/server/agent.cpp
```diff
@@ -4,231 +4,416 @@
...
+        }
+
+        // remove command from active command list.
+        active_commands_->Delete(cmd_id);
+
+        return ClientState::ACTIVE;
+
+    }
+
+    Agent::ClientState Agent::HandleRequest(
```
Nickster258, line 368: The current client code doesn't ever send a `QuitRequest`, though it probably should.

Either this is unnecessary, or the client should be telling the server it is disconnecting. I think the later makes more sense.

project1/server/agent.cpp
```diff
@@ -4,231 +4,416 @@
...
+        std::string cmd_str = request.command_id();
+
+        // Case to uint16_t
+        int cmd_int(std::stoi(cmd_str));
+        uint16_t cmd_id(0);
+
+        if (cmd_int <= static_cast<int>(UINT16_MAX) && cmd_int >=0) {
+            cmd_id = static_cast<uint16_t>(cmd_int);
+        }
+        else {
```
Nickster258: (minor) Accidental newline here maybe?

Nickster258 (changes requested):
I only really have a couple of smaller "food for thought ?" comments. I took specific note to how the agent handles the GET and PUT communications, and that looks like it should work as expected.

project1/server/agent.cpp
jake-chandler: It does seem redundant, but I think it's necessary to have two seperate, because, during `PUT` requests, we have to first recieve a `PutRequest` , send a `PutResponse`, and then explicitly wait for a `FileContents Message`.  These methods also use two different `MessageParser` objects. For these reasons I'm currently not seeing an intuitive way to combine these functions. 

As far as breaking out of the loop into a new funciton, I'd only be able to encapsulate a small amount of code into it, so that's not even worth it.

project1/server/agent.cpp
```diff
@@ -185,50 +180,19 @@ namespace server {
...
-        uint16_t totalBytesSent = 0;
-
-        // make 1000 byte buffer
-        int numBytesToSend = 1000;
-        uint8_t buf[numBytesToSend];
-
-        // counter used to track
-        int byte_counter = 0;
-
+        uint16_t kBytesSent = 0;
```
djpetti: (minor) If this is not a constant, don't name it like one.

project1/server/agent.cpp
```diff
@@ -185,50 +180,19 @@ namespace server {
...
-
-                    // update
-                    totalBytesSent += 1000;
-                }
-                buf[byte_counter] = outgoing_message_buffer_.at(n);
-                ++byte_counter;
-
+        while (kBytesSent < outgoing_message_buffer_.size()) {
+            if (!terminated) {
+                kBytesSent += send(client_fd_, outgoing_message_buffer_.data() + kBytesSent,
```
djpetti: Be careful with this. `send` can return -1 on failure, so you should probably explicitly check that.

djpetti: There is a bug here if `outgoing_message_buffer_.size()` is not a perfect multiple of 1000, in which case it will try to read past the end of the buffer on the last message. I think in the code snippet I posted about this, I had some logic using `std::min` to handle this case.

project1/server/agent.cpp
```diff
@@ -392,10 +346,9 @@ namespace server {
         int cmd_int(std::stoi(cmd_str));
         uint16_t cmd_id(0);
 
-        if (cmd_int <= static_cast<int>(UINT16_MAX) && cmd_int >=0) {
+        if (cmd_int <= static_cast<int>(UINT16_MAX) && cmd_int >= 0) {
```
djpetti: Prefer `std::numeric_limits<uint16_t>::max()` from `<limits>` instead.

project1/server/agent.h
```diff
@@ -127,6 +123,12 @@ namespace server {
          */
         bool SendFileContents(const ftp_messages::FileContents &file_contents, uint16_t command_id);
 
+        /**
+         * @bried Generates a command ID for a GET or PUT request.
```
djpetti: spelling

project1/server/server.cpp
```diff
@@ -1,41 +1,31 @@
...
-        tPortTask->SetPort(tPort);
+        auto nPortTask = std::make_shared<server_tasks::NPortTask>(active_ids,nPort,
+                                                                   read_manager_,write_manager_);
+        auto tPortTask = std::make_shared<server_tasks::TPortTask>(active_ids,tPort,
+                                                                   read_manager_,write_manager_);
 
         pool.AddTask(nPortTask);
         pool.AddTask(tPortTask);
 
+        while (pool.GetTaskStatus(nPortTask) != thread_pool::Task::Status::DONE ||
```
djpetti: I think you technically want to check `status == thread_pool::Task::Status::RUNNING` instead. The current logic would not handle the `FAILURE` case correctly.

project1/server/server_tasks/agent_task.h
```diff
@@ -14,38 +14,27 @@
...
-        thread_pool::Task::Status SetUp();
-        void CleanUp() final;
 
-        /**
-         * @brief Sets the ClientFD for this agent.
-         * @param id
-         * @note Needs to be set before being added to the thread pool.
-         */
-        void SetClientFD(int id);
+        AgentTask(int id, std::shared_ptr<CommandIDs> cmds,
```
djpetti, line 18: These constructors need docs.

project1/server/server_tasks/agent_task.cpp
```diff
@@ -1,25 +1,30 @@
...
     thread_pool::Task::Status AgentTask::SetUp() {
-        // give the agent a unique file handler with the shared access managers
-        auto fh = std::make_unique<server::file_handler::ThreadSafeFileHandler>(read_manager_,write_manager_);
-        agent_ = new server::Agent(client_fd_, std::move(fh),active_commands_);
+
+        // if this is an agent for a normal command, these members should be initialized
+        if (read_manager_ && write_manager_) {
+            // give the agent a unique file handler with the shared access managers
+            auto fh = std::make_unique<server::file_handler::ThreadSafeFileHandler>(read_manager_, write_manager_);
+            agent_ = new server::Agent(client_fd_, std::move(fh), active_commands_);
```
djpetti: Don't use `new` and `delete`. Use `std::unique_ptr` instead.

project1/server/server_tasks/command_ids.cpp
```diff
@@ -2,20 +2,17 @@
...
     }
     void CommandIDs::Delete(uint16_t id) {
         std::lock_guard<std::mutex> guard(mutex_);
-        command_ids_.remove(id);
+        command_ids_.erase(id);
     }
     bool CommandIDs::Contains(uint16_t id) {
         std::lock_guard<std::mutex> guard(mutex_);
-        std::list<uint16_t>::iterator find = std::find(command_ids_.begin(),
+        std::unordered_set<uint16_t>::iterator find = std::find(command_ids_.begin(),
```
djpetti: I think you can just do this with `const auto& find = command_ids_.find(id);`

djpetti (changes requested):
Still found one bug here.

project1/server/server_tasks/agent_task.cpp
jake-chandler: Apologies for not fixing this last time, I had some trouble figuring out syntax.

project1/server/agent.h
jake-chandler: This one made me chuckle,

project1/server/server_tasks/agent_task.cpp
jake-chandler: Got it.

project1/server/agent.cpp
```diff
@@ -180,15 +182,18 @@ namespace server {
         }
 
         // monitor for termination request for get and put commands
-        uint16_t kBytesSent = 0;
-        const uint32_t kBytesToSend = 1000;
+        uint16_t total_bytes_sent = 0;
```
djpetti: Probably want this not to be a `uint16_t`. If they're sending 5 MB files, this is going to overflow.

project1/server/agent.cpp
```diff
@@ -180,15 +182,18 @@ namespace server {
...
         bool terminated = false;
 
         // continue until we've sent the entire message
-        while (kBytesSent < outgoing_message_buffer_.size()) {
+        while (total_bytes_sent < outgoing_message_buffer_.size()) {
             if (!terminated) {
-                kBytesSent += send(client_fd_, outgoing_message_buffer_.data() + kBytesSent,
-                                   kBytesToSend, 0);
+                const uint32_t kBytesToSend = std::min( (int) outgoing_message_buffer_.size() - total_bytes_sent, 1000);
```
djpetti: This should be compared to `kBytesToSend`, instead of directly comparing to 1000. While we're at it, `kBytesToSend` is probably not the best name for this constant. I would suggest `kSendChunkSize` or something.

project1/server/server_tasks/agent_task.cpp
```diff
@@ -19,10 +19,10 @@ namespace server_tasks {
         if (read_manager_ && write_manager_) {
             // give the agent a unique file handler with the shared access managers
             auto fh = std::make_unique<server::file_handler::ThreadSafeFileHandler>(read_manager_, write_manager_);
-            agent_ = new server::Agent(client_fd_, std::move(fh), active_commands_);
+            agent_ = std::make_unique<server::Agent>(client_fd_,std::move(fh),std::move(active_commands_));
```
djpetti: I'm honestly not sure what `std::move`ing a shared pointer does in this context, but it doesn't make a lot of sense in this case. I think you should just pass it normally. (For `fh` it is valid.)

project1/server/server_tasks/agent_task.cpp
```diff
@@ -19,10 +19,10 @@ namespace server_tasks {
...
         } else {
             // no need to give an Agent a file handler for termination commands
-            agent_ = new server::Agent(client_fd_, active_commands_);
+            agent_ = std::make_unique<server::Agent>(client_fd_,std::move(active_commands_));
```
djpetti: ditto

djpetti (changes requested):
There is still at least one minor issue here.

project1/server/server_tasks/agent_task.cpp
jake-chandler: There is actually an error when I pass only `fh`...

project1/server/agent.cpp
djpetti: I agree. The client code should really send this request when the user types "quit". I think that behavior is actually a requirement for the project (or it was at least a requirement for project 1).

project1/server/agent.cpp
```diff
@@ -182,15 +182,15 @@ namespace server {
...
         // monitor for termination request for get and put commands
-        uint16_t total_bytes_sent = 0;
+        uint32_t total_bytes_sent = 0;
         bool terminated = false;
 
         // continue until we've sent the entire message
         while (total_bytes_sent < outgoing_message_buffer_.size()) {
             if (!terminated) {
-                const uint32_t kBytesToSend = std::min( (int) outgoing_message_buffer_.size() - total_bytes_sent, 1000);
+                const uint32_t kSendChunkSize = std::min((int) outgoing_message_buffer_.size() - total_bytes_sent, kSendChunkSize);
```
djpetti: Ah, whoops, sorry. I got confused by the variable naming. `kBytesToSend` is a fine name for the result of the `std::min` call. I guess what I meant to suggest is that you add a new `constexpr` variable called `kChunkSize` which is set to 1000, kind of like what you did on the receiving side. That would be your second argument to `std::min`.

project1/server/server.cpp
```diff
@@ -14,16 +14,16 @@ namespace server {
         std::shared_ptr<server_tasks::CommandIDs> active_ids;
 
         // pass active command list to nPortTask and tPortTask
-        auto nPortTask = std::make_shared<server_tasks::NPortTask>(active_ids,nPort,
-                                                                   read_manager_,write_manager_);
-        auto tPortTask = std::make_shared<server_tasks::TPortTask>(active_ids,tPort,
-                                                                   read_manager_,write_manager_);
+        auto nPortTask = std::make_shared<server_tasks::NPortTask>(std::move(active_ids),nPort,
+                                                                   std::move(read_manager_),std::move(write_manager_));
```
djpetti: Again, I don't think `std::move`ing shared pointer arguments gets you anything. The confusion might be stemming from the fact that it is necessary for `unique_ptr`, but `shared_ptr` has different semantics in that regard. It might be worth reading about C++ move semantics if you're curious.

project1/server/server.cpp
```diff
@@ -14,16 +14,16 @@ namespace server {
...
+        auto tPortTask = std::make_shared<server_tasks::TPortTask>(std::move(active_ids),tPort,
+                                                                   std::move(read_manager_),std::move(write_manager_));
 
         pool.AddTask(nPortTask);
         pool.AddTask(tPortTask);
 
-        while (pool.GetTaskStatus(nPortTask) != thread_pool::Task::Status::DONE ||
-               pool.GetTaskStatus(tPortTask) != thread_pool::Task::Status::DONE) {
+        while (pool.GetTaskStatus(nPortTask) != thread_pool::Task::Status::RUNNING ||
```
djpetti: Wrong logic here. This should be `==` now, not `!=`.

project1/server/server_tasks/agent_task.cpp
```diff
@@ -18,11 +18,11 @@ namespace server_tasks {
         // if this is an agent for a normal command, these members should be initialized
         if (read_manager_ && write_manager_) {
             // give the agent a unique file handler with the shared access managers
-            auto fh = std::make_unique<server::file_handler::ThreadSafeFileHandler>(read_manager_, write_manager_);
-            agent_ = std::make_unique<server::Agent>(client_fd_,std::move(fh),std::move(active_commands_));
+            auto fh = std::make_unique<server::file_handler::ThreadSafeFileHandler>(std::move(read_manager_), std::move(write_manager_));
```
djpetti, line 21: Same comment about `std::move`.

project1/server/server_tasks/nport_task.cpp
```diff
@@ -24,8 +24,8 @@ namespace server_tasks {
 
             std::cout << "Handling new connection from client #" << client_fd << "." << std::endl;
 
-            auto agent_task = std::make_shared<AgentTask>(client_fd, active_ids_,
-                                                          read_manager_, write_manager_);
+            auto agent_task = std::make_shared<AgentTask>(client_fd, std::move(active_ids_),
```
djpetti: And... here too.

project1/server/server_tasks/tport_task.cpp
```diff
@@ -22,7 +22,7 @@ namespace server_tasks {
 
             std::cout << "Handling new connection from client #" << client_fd << "." << std::endl;
 
-            auto agent_task = std::make_shared<AgentTask>(client_fd, active_ids_);
+            auto agent_task = std::make_shared<AgentTask>(client_fd, std::move(active_ids_));
```
djpetti: ditto

djpetti (changes requested):
Well, there's still a bug.

project1/server/server.cpp
jake-chandler: i definitely was getting `unique_ptr` and `shared_ptr` confused, mainly becuase I'm not exactly sure how their functionality differs.

project1/server/server.cpp
jake-chandler: Duh. Whoops.

djpetti (approved):
Looks good... finally!

Nickster258 (approved):

PR #10, opened 2021-03-08
Add ThreadPool support for waiting on task completion.

djpetti:
This should make it easier to do things like idle the main thread until all tasks are complete.


jake-chandler (approved):
This functionality is much appreciated.

Nickster258:
> This functionality is much appreciated.

Agreed. This could come in really handy.

PR #11, opened 2021-03-08
Add forking, terminate support to client

Nickster258:
This adds forking support to the GET and PUT commands.

By default, the input parser will recognize any command as forking if the last word of the command is `&`. However, the client is only programmed to parse these commands as forked processes for GET and PUT commands.

Note: I am going to add logging support in a later commit, just want this out there for review.

Also, there may be unforeseen issues with how I handled passing variables off to the tasks, so please take specific note to that.

### DIFF
--- a/project1/client/client.cpp
+++ b/project1/client/client.cpp
@@ -1,4 +1,6 @@
 #include "client.h"
+
+#include <memory>
 #include "client_tasks/terminate_task.h"
 #include "client_tasks/upload_task.h"
 #include "client_tasks/download_task.h"
@@ -122,7 +124,7 @@ void Client::FtpShell() {
     std::getline(std::cin, input);
 
     // determine command
-    auto ip_ = new InputParser(input);
+    std::unique_ptr<input_parser::InputParser> ip_ = std::make_unique<input_parser::InputParser>(input);
 
     // create & serialize request message for determined command
     r = ip_->CreateReq();
@@ -133,13 +135,11 @@ void Client::FtpShell() {
     if (r.has_terminate()) {
       auto terminate_task = std::make_shared<client_tasks::TerminateTask>(hostname_, tport_, r.terminate());
       pool.AddTask(terminate_task);
-      delete ip_;
       continue;
     }
     if (r.has_quit()) {
       // close socket connection
       close(client_fd_);
-      delete ip_;
       // exit loop, no need to send request
       connected_ = false;
       continue;
@@ -170,7 +170,6 @@ void Client::FtpShell() {
         pool.AddTask(get_task);
       }
     }
-    delete ip_;
   }
 }
 }  // namespace client
--- a/project1/client/client.cpp
+++ b/project1/client/client.cpp
@@ -114,7 +114,7 @@ bool Client::FtpShell() {
   thread_pool::ThreadPool pool;
 
   while (connected_) {
-    // will have to pay attention to formatting here when implementing Output()
+    // display myftp prompt
     std::cout << "\nmyftp> ";
 
     // reset input string
--- a/project1/client/client.cpp
+++ b/project1/client/client.cpp
@@ -151,6 +151,7 @@ bool Client::FtpShell() {
     WaitForMessage();
     HandleResponse();
     if (r.has_put()) {
+      // If put command, FileContents will have to be sent.
       auto contents = ip_->GetContentsMessage();
       wire_protocol::Serialize(contents, &outgoing_msg_buf_);
       if (!ip_->IsForking()) {
@@ -160,6 +161,7 @@ bool Client::FtpShell() {
         pool.AddTask(put_task);
       }
     } else if (r.has_get()) {
+      // If get command, FileContents will have to be received.
       if (!ip_->IsForking()) {
         WaitForMessage();
         client_util::SaveIncomingFile(parser_, ip_->GetFilename());
--- a/project1/client/client.cpp
+++ b/project1/client/client.cpp
@@ -1,10 +1,8 @@
 #include "client.h"
 #include "client_tasks/terminate_task.h"
-
-#include <netinet/in.h>
-#include <sys/socket.h>
-
-#include <iostream>
+#include "client_tasks/upload_task.h"
+#include "client_tasks/download_task.h"
+#include "../thread_pool/thread_pool.h"
 
 namespace client {
 
@@ -15,7 +13,7 @@ bool Client::Connect(const std::string &hostname, uint16_t nport, uint16_t tport
   hostname_ = hostname;
   nport_ = nport;
   tport_ = tport;
-  connected_ = true;
+  connected_ = client_fd_ != -1;
   return connected_;
 }
 
@@ -35,8 +33,8 @@ bool Client::SendReq() {
   return true;
 }
 
-bool Client::WaitForResponse() {
-  while (!response_parser_.HasCompleteMessage()) {
+bool Client::WaitForMessage() {
+  while (!parser_.HasCompleteMessage()) {
     incoming_msg_buf_.resize(kBufferSize);
 
     const auto bytes_read =
@@ -49,7 +47,7 @@ bool Client::WaitForResponse() {
 
     incoming_msg_buf_.resize(bytes_read);
 
-    response_parser_.AddNewData(incoming_msg_buf_);
+    parser_.AddNewData(incoming_msg_buf_);
   }
 
   // empty buffer
@@ -63,7 +61,7 @@ void Client::HandleResponse() {
 
   ftp_messages::Response msg;
 
-  response_parser_.GetMessage(&msg);
+  parser_.GetMessage(&msg);
 
   // determine type of response and respond accordingly
   if (msg.has_list()) {
@@ -79,20 +77,18 @@ void Client::HandleResponse() {
   } else if (msg.has_pwd()) {
     auto pwd_response = msg.pwd();
     output_ = pwd_response.dir_name();
+  } else if (msg.has_put()) {
+    auto put_response = msg.put();
+    output_ = put_response.command_id();
+  } else if (msg.has_get()) {
+    auto get_response = msg.get();
+    output_ = get_response.command_id();
   }
 
   // make sure outputting is necessary
-  if (output_.compare("")) {
+  if (!output_.empty()) {
     Output();
   }
-}
-
-void Client::HandleFileContents() {
-  ftp_messages::FileContents contents;
-  file_parser_.GetMessage(&contents);
-  std::ofstream new_file;
-  new_file.open(ip_->GetFilename());
-  new_file << contents.contents();
 }
 
 void Client::Output() {
@@ -139,6 +135,7 @@ bool Client::FtpShell() {
     if (r.has_terminate()) {
       auto terminate_task = std::make_shared<client_tasks::TerminateTask>(hostname_, tport_, r.terminate());
       pool.AddTask(terminate_task);
+      delete ip_;
       continue;
     }
     if (r.has_quit()) {
@@ -151,18 +148,25 @@ bool Client::FtpShell() {
     }
     wire_protocol::Serialize(r, &outgoing_msg_buf_);
     SendReq();
-    WaitForResponse();
-    if (r.has_get() || r.has_put()) {
-      // Since these two requests are the only two that require more messages,
-      // Continue processing messages accordingly
+    WaitForMessage();
+    HandleResponse();
+    if (r.has_put()) {
+      auto contents = ip_->GetContentsMessage();
+      wire_protocol::Serialize(contents, &outgoing_msg_buf_);
       if (!ip_->IsForking()) {
-        HandleResponse();
+        SendReq();
       } else {
-        // TODO forked tasks will be handled here
+        auto put_task = std::make_shared<client_tasks::UploadTask>(client_fd_, outgoing_msg_buf_);
+        pool.AddTask(put_task);
       }
-      continue;
-    } else {
-      HandleResponse();
+    } else if (r.has_get()) {
+      if (!ip_->IsForking()) {
+        WaitForMessage();
+        client_util::SaveIncomingFile(parser_, ip_->GetFilename());
+      } else {
+        auto get_task = std::make_shared<client_tasks::DownloadTask>(ip_->GetFilename(), incoming_msg_buf_, kBufferSize, client_fd_, parser_);
+        pool.AddTask(get_task);
+      }
     }
     delete ip_;
   }
--- a/project1/client/client.cpp
+++ b/project1/client/client.cpp
@@ -124,11 +124,11 @@ void Client::FtpShell() {
     std::getline(std::cin, input);
 
     // determine command
-    std::unique_ptr<input_parser::InputParser> ip_ = std::make_unique<input_parser::InputParser>(input);
+    std::unique_ptr<input_parser::InputParser> ip = std::make_unique<input_parser::InputParser>(input);
 
     // create & serialize request message for determined command
-    r = ip_->CreateReq();
-    if (!ip_->IsValid()) {
+    r = ip->CreateReq();
+    if (!ip->IsValid()) {
       std::cout << "Invalid command, try again." << "\n";
       continue;
     }
@@ -150,9 +150,9 @@ void Client::FtpShell() {
     HandleResponse();
     if (r.has_put()) {
       // If put command, FileContents will have to be sent.
-      auto contents = ip_->GetContentsMessage();
+      auto contents = ip->GetContentsMessage();
       wire_protocol::Serialize(contents, &outgoing_msg_buf_);
-      if (!ip_->IsForking()) {
+      if (!ip->IsForking()) {
         SendReq();
       } else {
         auto put_task = std::make_shared<client_tasks::UploadTask>(client_fd_);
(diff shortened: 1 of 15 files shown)


project1/client/client.cpp
```diff
@@ -158,14 +111,15 @@ void Client::Output() {
 }
 bool Client::FtpShell() {
   ftp_messages::Request r;
+  thread_pool::ThreadPool pool;
 
   while (connected_) {
-    // reset input string
-    std::string input;
-
     // will have to pay attention to formatting here when implementing Output()
```
jake-chandler: Please remove this comment. I meant to remove it before submitting p1 but a clearly didn't...

jake-chandler: Please remove this comment. I meant to remove it before submitting p1 but a clearly didn't...

project1/client/input_parser/input_parser.cpp
```diff
@@ -9,64 +9,63 @@ using ftp_messages::MakeDirRequest;
...
 
 std::string InputParser::GetFilename() { return fn_; }
-Request InputParser::CreateGetReq() {
-  Request request;
-  request.mutable_get()->set_filename(fn_);
-  return request;
+bool InputParser::IsValid() { return is_valid_; }
+bool InputParser::IsForking() { return is_forking_; }
+
+FileContents InputParser::GetContentsMessage() {
```
jake-chandler, line 64: What exactly is the purpose of this? I can't seem to figure out what you're doing here.

project1/client/client.cpp
```diff
@@ -174,21 +128,49 @@ bool Client::FtpShell() {
...
-    wire_protocol::Serialize(r, &outgoing_msg_buf_);
-
+    if (!ip_->IsValid()) {
+      std::cout << "Invalid command, try again." << "\n";
+      continue;
+    }
+    if (r.has_terminate()) {
+      auto terminate_task = std::make_shared<client_tasks::TerminateTask>(hostname_, tport_, r.terminate());
+      pool.AddTask(terminate_task);
+      delete ip_;
```
djpetti: Much better to use a `unique_ptr` to store this instead of manually managing memory.

project1/client/input_parser/input_parser.cpp
```diff
@@ -9,64 +9,63 @@ using ftp_messages::MakeDirRequest;
...
-            req_ = QUIT;
-            break;
-        }
-      }
-    }
-  } while (iss);
+  std::string word;
+  server::file_handler::FileHandler fh;
+  iss >> word;
+  auto itr = commands_.find(word);
```
jake-chandler, line 26: This is much better than what I had before, nice job.

project1/client/client.cpp
```diff
@@ -174,21 +128,49 @@ bool Client::FtpShell() {
...
+    WaitForMessage();
     HandleResponse();
+    if (r.has_put()) {
+      // If put command, FileContents will have to be sent.
+      auto contents = ip_->GetContentsMessage();
+      wire_protocol::Serialize(contents, &outgoing_msg_buf_);
+      if (!ip_->IsForking()) {
+        SendReq();
+      } else {
+        auto put_task = std::make_shared<client_tasks::UploadTask>(client_fd_, outgoing_msg_buf_);
```
djpetti: Is `outgoing_msg_buf_` thread-safe? If not, I think each individual `UploadTask` is going to have to create their own instead of receiving it from the parent.

project1/client/client.cpp
```diff
@@ -174,21 +128,49 @@ bool Client::FtpShell() {
...
+        pool.AddTask(put_task);
+      }
+    } else if (r.has_get()) {
+      // If get command, FileContents will have to be received.
+      if (!ip_->IsForking()) {
+        WaitForMessage();
+        client_util::SaveIncomingFile(parser_, ip_->GetFilename());
+      } else {
+        auto get_task = std::make_shared<client_tasks::DownloadTask>(ip_->GetFilename(), incoming_msg_buf_, kBufferSize, client_fd_, parser_);
```
djpetti: Same thread-safety comment about `incoming_msg_buf_`.

project1/client/client.h
```diff
@@ -5,87 +5,93 @@
...
+ *
+ */
+class Client {
+ public:
+  /**
+   * @brief The shell logic.
+   *
+   * @return True when user properly quits
+   *
+   * @note will never return false.
```
djpetti, line 29: Why not make this `void` then?

project1/client/client.h
```diff
@@ -5,87 +5,93 @@
...
+   */
+  bool FtpShell();
+
+  /**
+   *
+   * @param hostname the ip address of the FTP server
+   * @param port the port the FTP server is binded to
```
djpetti: Update docs with new arguments.

project1/client/client_tasks/download_task.cpp
```diff
@@ -0,0 +1,18 @@
+#include "download_task.h"
+
+namespace client_tasks {
+thread_pool::Task::Status DownloadTask::RunAtomic() {
+  while(!parser_.HasCompleteMessage()) {
+    incoming_file_buf_.resize(buffer_size_);
+    recv(client_fd_, incoming_file_buf_.data(), buffer_size_, 0);
+    incoming_file_buf_.resize(buffer_size_);
```
djpetti: I think what you mean to do is check the return value of `recv` and resize the buffer to this value. You probably should do that anyway for error-handling purposes...

jake-chandler (commented):
Add documentation to most of the `.h` files. Otherwise, everything seems to be logically intact. I tried to pay _extra_ attention to how you handled forking on the `InputParser` , and I believe it fits according to the project description. As far as passing variables to tasks, I had similar concerns. I'll wait to approve until Daniel reviews.

project1/client/client_tasks/download_task.h
```diff
@@ -0,0 +1,34 @@
+#ifndef PROJECT1_DOWNLOAD_TASK_H
```
djpetti, line 5: Missing docs in this file.

project1/client/client_tasks/download_task.h
```diff
@@ -0,0 +1,34 @@
...
+#define PROJECT1_DOWNLOAD_TASK_H
+
+#include "../../thread_pool/task.h"
+#include "../../wire_protocol/wire_protocol.h"
+#include "../client_util.h"
+
+namespace client_tasks {
+class DownloadTask : public thread_pool::Task {
+ public:
+  DownloadTask(const std::string& filename, std::vector<uint8_t> &incoming_buf,
```
djpetti: (minor) Generally good to avoid non-const references as parameters, because it's not obvious to the callee that the parameters can be modified. If these parameters cannot be const, pass them by pointer instead.

project1/client/client_tasks/download_task.h
```diff
@@ -0,0 +1,34 @@
...
+               size_t buf_size, int client_fd,
+               wire_protocol::MessageParser<google::protobuf::Message> &parser
+               ) {
```
djpetti: (minor) Prefer initialization list syntax to initialize class members. Also, any reason this implementation isn't in the .cpp file?

project1/client/client_tasks/download_task.h
```diff
@@ -0,0 +1,34 @@
...
+  Status RunAtomic() override;
+
+  void CleanUp() override;
+
+ protected:
+  std::string filename_{};
+  std::vector<uint8_t> incoming_file_buf_{};
+  int client_fd_;
+  size_t buffer_size_;
+  wire_protocol::MessageParser<google::protobuf::Message> parser_;
```
djpetti: A message parser that merely parses generic messages is probably not going to be useful to you. Since you're only parsing `FileContents` messages, you probably want to use that as the type parameter for the parser. Furthermore, `MessageParser` isn't thread-safe anyway, so you really should be creating a new one for this class instead of accepting one from the parent.

project1/client/client_tasks/terminate_task.cpp
```diff
@@ -0,0 +1,21 @@
...
+}
+thread_pool::Task::Status TerminateTask::RunAtomic() {
+  wire_protocol::Serialize(terminate_req_, &outgoing_terminate_buf_);
+  if (send(socket_, outgoing_terminate_buf_.data(),
+           outgoing_terminate_buf_.size(), 0) < 0) {
+    perror("Failed to send request");
+    return thread_pool::Task::Status::FAILED;
+  }
+  outgoing_terminate_buf_.clear();
+  return thread_pool::Task::Status::RUNNING;
```
djpetti: Did you mean to return `DONE` here? If you return `RUNNING`, it will keep calling this method over and over again.

project1/client/input_parser/input_parser.cpp
Nickster258: Thanks! I admittedly spent too long trying to directly have an enum with strings as the command for each key, but had to settle for the map.

project1/client/client_tasks/terminate_task.h
```diff
@@ -0,0 +1,37 @@
...
+#include "../../wire_protocol/wire_protocol.h"
+#include "ftp_messages.pb.h"
+#include "../client_util.h"
+
+#include <sys/socket.h>
+#include <utility>
+#include <unistd.h>
+
+namespace client_tasks {
+class TerminateTask : public thread_pool::Task {
```
djpetti, line 18: Broadly, I question whether a separate thread is necessary for this. When a terminate request is sent, it seems like all it has to do is connect to a socket and send a short message to the server, which should be quite fast and can therefore probably be done in the main thread.

project1/client/client_tasks/terminate_task.h
```diff
@@ -0,0 +1,37 @@
+#ifndef PROJECT1_TERMINATE_TASK_H
```
djpetti, line 5: Missing docs in this file.

project1/client/client_tasks/upload_task.h
```diff
@@ -0,0 +1,24 @@
+#ifndef PROJECT1_UPLOAD_TASK_H
```
djpetti, line 5: Missing docs for this file.

project1/client/client_util.h
```diff
@@ -0,0 +1,54 @@
...
+#include <netinet/in.h>
+#include <sys/socket.h>
+#include <arpa/inet.h>
+#include <fstream>
+
+#include "ftp_messages.pb.h"
+#include "../wire_protocol/wire_protocol.h"
+
+namespace client_util {
+inline struct sockaddr_in MakeAddress(uint16_t port) {
```
djpetti, line 16: I generally avoid `inline` functions. Usually, the compiler is smart enough to auto-inline code for you anyway when it thinks it can, making this keyword effectively meaningless. To that end, is there any reason you didn't put the definitions for these in a .cpp file?

project1/client/client_util.h
```diff
@@ -0,0 +1,54 @@
...
+  return sock;
+}
+inline void SaveIncomingFile(
+    wire_protocol::MessageParser<google::protobuf::Message> f_parser,
+    const std::string& name) {
+  ftp_messages::FileContents contents;
+  f_parser.GetMessage(&contents);
+  std::ofstream new_file;
+  new_file.open(name);
+  new_file << contents.contents();
```
djpetti: I'm really surprised this compiles, considering that a plain `google::protobuf::Message` should not have a `contents` method.

djpetti (changes requested):
This was also a first pass. I'll look again once you're finished with these comments.

Nickster258:
> Add documentation to most of the `.h` files. Otherwise, everything seems to be logically intact. I tried to pay _extra_ attention to how you handled forking on the `InputParser` , and I believe it fits according to the project description. As far as passing variables to tasks, I had similar concerns. I'll wait to approve until Daniel reviews.

Resolved in 69221518e647b436755b5e0a2d43231531da6317

project1/client/client.cpp
Nickster258: I added this in 06164ccfeae0a8ced9aa0f3638868120c13f076d. I believe I did it correctly, but will need to be given a quick glance.

project1/client/client.cpp
Nickster258: It likely is not, so I made a new one for each `UploadTask` as you recommended.

project1/client/client_util.h
Nickster258: I actually did that because I couldn't get the `client_util.cpp` file to behave properly. I am bound to be doing something wrong though.

project1/client/client_tasks/download_task.h
Nickster258: Created a new parser looking for the specific `FileContents` type.

project1/client/client.cpp
djpetti: Noted. I'll look.

project1/client/client_util.h
djpetti: Interesting... What specific issue were you having?

project1/client/client_tasks/download_task.h
Nickster258: Can you expand more on what you mean by list syntax? I'm not familiar.

I did not know where best instantiate this. Should it be in the .cpp?

project1/client/client.cpp
```diff
@@ -124,7 +124,7 @@ bool Client::FtpShell() {
     std::getline(std::cin, input);
 
     // determine command
-    ip_ = new InputParser(input);
+    std::unique_ptr<input_parser::InputParser> ip_ = std::make_unique<input_parser::InputParser>(input);
```
djpetti: This looks fine to me. Only thing is, I'd remove the trailing underscore from the name since it's a local variable and not a class member.

project1/client/client_tasks/download_task.h
```diff
@@ -8,27 +12,32 @@
...
-    parser_ = parser;
   }
 
   Status RunAtomic() override;
 
   void CleanUp() override;
 
  protected:
+
+  // the name of the file to be retreived
```
djpetti: (minor) Single-line documentation comments should have triple slashes. This is so they will be interpreted as such if we ever decide to run Doxygen. (admittedly unlikely)

project1/client/client_tasks/terminate_task.h
Nickster258: I just made it a separate thread out of convenience. It was honestly a personal preference to separate what happens on the main thread and nport from tport requests.

Should we put it into the main thread?

project1/client/client_tasks/download_task.h
djpetti: I'm referring to [this](https://en.cppreference.com/w/cpp/language/constructor). In the way of examples, pretty much all the other classes in this repo that take constructor arguments use this strategy. I usually do put it in the .cpp file.

project1/client/client_tasks/terminate_task.h
djpetti: Umm, it's not a huge deal at this point. My very general suggestion is to avoid concurrency whenever possible because it tends to lead to difficult bugs.

djpetti (changes requested):
Responding to some comments...

project1/client/client_tasks/download_task.cpp
```diff
@@ -1,6 +1,12 @@
 #include "download_task.h"
 
 namespace client_tasks {
+DownloadTask::DownloadTask(const std::string &filename, size_t buf_size,
```
djpetti: I take it you decided not to attempt an initialization list?

For reference, that would look like
```c++
DownloadTask::DownloadTask(const std::string &filename, size_t buf_size, 
                                                  int client_fd) : filename_(filename), buffer_size_(buf_size),
                                                                        client_fd_(client_fd) {}
```

I apologize for the weird formatting; I have to do it by hand.

project1/client/client_tasks/download_task.h
```diff
@@ -12,31 +12,27 @@
 namespace client_tasks {
 class DownloadTask : public thread_pool::Task {
  public:
-  DownloadTask(const std::string& filename, size_t buf_size, int client_fd) {
-    filename_ = filename;
-    buffer_size_ = buf_size;
-    client_fd_ = client_fd;
-  }
+  DownloadTask(const std::string& filename, size_t buf_size, int client_fd);
```
djpetti: The constructor should probably have a docstring documenting the parameters.

project1/client/client_tasks/terminate_task.h
```diff
@@ -18,11 +18,7 @@ namespace client_tasks {
 class TerminateTask : public thread_pool::Task {
  public:
   TerminateTask(const std::string &address, uint16_t port,
```
djpetti, line 25: Same comments on this constructor.

project1/client/client_tasks/upload_task.h
```diff
@@ -12,19 +12,18 @@
 namespace client_tasks {
 class UploadTask : public thread_pool::Task {
  public:
-  UploadTask(int client_fd) {
-    client_fd_ = client_fd;
-  }
+  UploadTask(int client_fd);
```
djpetti: ...and this one. Also, this should probably be marked `explicit`? I thought CLion usually yelled at you if you forgot to do that.

djpetti (approved):
Once comments are resolved, this is good to merge.

project1/client/client_tasks/download_task.cpp
Nickster258: Oh I see. I didn't catch that before. That is a convenient thing to use. I am not sure of the best way to style it, but we can figure that out later.

project1/client/client_tasks/download_task.h
Nickster258: Admittedly, it was initially const so I could get the code to compile and function like I wanted.

PR #14, opened 2021-03-24
Reorganize Project Structure

djpetti:
This breaks out common functionality into a separate folder and paves the way for reusing it in multiple projects.


Nickster258:
I'm admittedly entirely new to proper C++/CMake project structure, but in browsing around examples online, this looks fairly straight forward and in-line with what I saw.

Nickster258 (approved):
I'm admittedly entirely new to proper C++/CMake project structure, but in browsing around examples online, this looks fairly straight forward and in-line with what I saw.

jake-chandler (approved):

PR #15, opened 2021-03-26
Add protobuf messages

djpetti:
I'm still not entirely clear on whether the participants actually have to listen on a port as opposed to just connecting to an open port the coordinator and receiving messages that way. If so, this will need to be changed.

### DIFF
--- a/project3/proto/pub_sub_messages.proto
+++ b/project3/proto/pub_sub_messages.proto
@@ -1,0 +1,57 @@
+syntax = "proto3";
+
+package pub_sub_messages;
+
+/// Request sent by a participant to register with the coordinator.
+message Register {
+}
+
+/// Request sent by the coordinator upon successful registration.
+message RegistrationResponse {
+  /// Indicates the ID of this participant.
+  uint32 participant_id = 1;
+}
+
+/// Message sent by a participant to deregister with the coordinator.
+message Deregister {
+}
+
+/// Message sent by a participant to indicate that it is going offline.
+message Disconnect {
+}
+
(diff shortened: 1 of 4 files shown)


project3/proto/pub_sub_messages.proto
```diff
@@ -0,0 +1,57 @@
...
+  /// The ID of the participant that this message originated from.
+  uint32 origin_id = 2;
+}
+
+/**
+ * @brief Encapsulates all messages that can be received by the coordinator. This is
+ * so we can simply parse this message type from the network and worry later about
+ * the actual message type.
+ */
+message CoordinatorMessage {
```
Nickster258, line 49: My only concern with this is that it encapsulates messages that the coordinator will be receiving, but the name implies it contains messages the coordinator will be sending. Should this be instead generalized for both, or possibly renamed to reflect project 1 and 2 structure?

Nickster258 (commented):
All in all it looks like these messages are all we would need for having a multicast server and client communicate. I only have one concern about the message encapsulation.

project3/proto/pub_sub_messages.proto
djpetti: I actually don't like the project1/2 structure as much. In practice, you really only need to do this encapsulation for messages that are being received, not send, hence I think it's better to name these encapsulation messages after the receiving end. I know that would certainly make more sense to me if I was skimming the coordinator code and came across a message of this type.

Nickster258: Yeah, that does actually make more sense in that case.

Nickster258 (approved):

PR #16, opened 2021-03-30
Add input_parser

Nickster258:
This is a simple implementation of input parser for Project 3. One main change between this and the original input parser is that this will not need to be re-instantiated. Instead, there is a method to parse new input and the request will be re-evaluated each time a message is parsed.

Note: I wanted to avoid any null-related issues with class variables so I added them all in the initializer list. However, I am not sure if this was the proper way to do it here.

### DIFF
--- a/project3/participant/input_parser/CMakeLists.txt
+++ b/project3/participant/input_parser/CMakeLists.txt
@@ -1,3 +1,2 @@
-add_library(input_parser_p3 input_parser.h input_parser.cpp)
-set_target_properties(input_parser_p3 PROPERTIES LINKER_LANGUAGE CXX)
+add_library(input_parser_p3 input_parser.cpp)
 target_link_libraries(input_parser_p3 pub_sub_proto wire_protocol)
--- a/project3/participant/input_parser/CMakeLists.txt
+++ b/project3/participant/input_parser/CMakeLists.txt
@@ -1,0 +1,3 @@
+add_library(input_parser_p3 input_parser.h input_parser.cpp)
+set_target_properties(input_parser_p3 PROPERTIES LINKER_LANGUAGE CXX)
+target_link_libraries(input_parser_p3 pub_sub_proto wire_protocol)
--- a/project3/participant/input_parser/input_parser.cpp
+++ b/project3/participant/input_parser/input_parser.cpp
@@ -3,11 +3,11 @@
 namespace participant::input_parser {
 
 InputParser::InputParser()
-    : reg_msg_{pub_sub_messages::Register()},
-      dereg_msg_{pub_sub_messages::Deregister()},
-      discon_msg_{pub_sub_messages::Disconnect()},
-      recon_msg_{pub_sub_messages::Reconnect()},
-      msend_msg_{pub_sub_messages::SendMulticast()},
+    : reg_msg_{},
+      dereg_msg_{},
+      discon_msg_{},
+      recon_msg_{},
+      msend_msg_{},
       multi_id_{0},
       is_valid_{true},
       req_{REG},
--- a/project3/participant/input_parser/input_parser.cpp
+++ b/project3/participant/input_parser/input_parser.cpp
@@ -3,7 +3,15 @@
 namespace participant::input_parser {
 
 InputParser::InputParser()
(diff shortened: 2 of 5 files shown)


project3/participant/input_parser/CMakeLists.txt
```diff
@@ -0,0 +1,3 @@
+add_library(input_parser_p3 input_parser.h input_parser.cpp)
```
djpetti: Don't need to include `.h` files here.

project3/participant/input_parser/CMakeLists.txt
```diff
@@ -0,0 +1,3 @@
...
+set_target_properties(input_parser_p3 PROPERTIES LINKER_LANGUAGE CXX)
```
djpetti: This line is *probably* unnecessary...

project3/participant/input_parser/input_parser.h
```diff
@@ -0,0 +1,87 @@
...
+                                                    {"reconnect", RECON},
+                                                    {"msend", MSEND}};
+
+  /// information to be extracted from input
+  int multi_id_;
+  bool is_valid_;
+  MsgType req_;
+  std::string message_;
+};
+};      // namespace participant::input_parser
```
djpetti: (minor) Namespaces don't need semicolons.

project3/participant/input_parser/input_parser.cpp
```diff
@@ -0,0 +1,65 @@
...
+      return CreateDisconMsg();
+    case RECON:
+      return CreateReconMsg();
+    case MSEND:
+      return CreateMsendMsg();
+  }
+  return nullptr;
+}
+pub_sub_messages::Register* InputParser::CreateRegMsg() {
+  return pub_sub_messages::Register().New();
```
djpetti: The Protobuf documentation is a little vague on this, but I think [messages created with `New()` have to be manually deleted](https://stackoverflow.com/questions/10217436/properly-freeing-protocol-buffer-memory/10217744). How about just creating an instance of each message type as class members and handing out pointers to those?

djpetti (changes requested):
One truly significant comment. Should be easy to fix.

project3/participant/input_parser/input_parser.cpp
```diff
@@ -3,7 +3,15 @@
 namespace participant::input_parser {
 
 InputParser::InputParser()
-    : multi_id_{0}, is_valid_{true}, req_{REG}, message_{} {}
+    : reg_msg_{pub_sub_messages::Register().New()},
```
djpetti: If you're going to use `New()`, you still need to wrap them in a `unique_ptr` so they get deleted correctly. Otherwise, you can just make them normal class members instead of pointers.

djpetti (changes requested):
I see you made changes, but you technically still have a memory leak.

jake-chandler (approved):

project3/participant/input_parser/input_parser.h
```diff
@@ -46,43 +46,18 @@ class InputParser {
...
                                                     {"reconnect", RECON},
                                                     {"msend", MSEND}};
 
   /// messages having been parsed from input
-  pub_sub_messages::Register* reg_msg_;
-  pub_sub_messages::Deregister* dereg_msg_;
-  pub_sub_messages::Disconnect* discon_msg_;
-  pub_sub_messages::Reconnect* recon_msg_;
-  pub_sub_messages::SendMulticast* msend_msg_;
+  pub_sub_messages::Register reg_msg_;
```
djpetti: Looks good! Just for reference, you should be able to initialize these members just by appending `{}` to the definition here. Then there will be no need to do it manually in the constructor. (Even if you don't, it should still work, but I think `clang-tidy` yells at you.)

djpetti (approved):
Minor organizational suggestion, but otherwise all good

project3/participant/input_parser/input_parser.h
Nickster258: Seems it didn't really yell at me interestingly enough. I do like how that cleaned up some of the initialization.

PR #25, opened 2021-04-02
Implement thread-safe queue

djpetti:
Adds a thread-safe queue data structure to the common library.


jake-chandler (approved):
Looks good. As always, the unit tests are appreciated

PR #27, opened 2021-04-03
Feature/messenger registrar api

jake-chandler:
The `Messenger` and `Registrar` classes are nearly complete. These will be used by the `Coordinator` class.


jake-chandler:
Do I need to make interactions with the socket connected to the participant's thread b thread-safe? It seems like there could be some concurrency issues with `BroadcastMessage()` and `SendMissedMessages()` in the `Messenger` class.

project3/coordinator/CMakeLists.txt
```diff
@@ -0,0 +1,2 @@
+add_library(coordinator connected_participants.cpp message_log.cpp registrar.cpp messenger.cpp)
+target_link_libraries(coordinator pub_sub_proto wire_protocol queue)
```
djpetti: I'm surprised this works, since there's no specific `queue` target. Since the queue is a header-only library, you don't have to specify any link libraries to use is.

project3/coordinator/connected_participants.h
```diff
@@ -0,0 +1,69 @@
+/**
+ * @file The connected Participants.
+ */
+#ifndef PROJECT3_CONNECTED_PARTICIPANTS_H
+#define PROJECT3_CONNECTED_PARTICIPANTS_H
+#include <algorithm>
+#include <mutex>
+#include <unordered_set>
+namespace coordinator::connected_participants {
```
djpetti: I think this is going a little overboard with the namespaces. I would just put everything in the `coordinator` namespace. That will make things more readable. In general, I use one namespace for every directory.

project3/coordinator/connected_participants.h
```diff
@@ -0,0 +1,69 @@
...
+
+  /**
+   * @brief Deletes a participant from the set of connected participants.
+   * @param participant
+   */
+  void DeleteParticipant(struct Participant& participant);
+
+  /**
+   * @brief Retrieves the set of currently connected participants.
+   * @note Will return a copy of the set, meant for read only.
```
djpetti: In this note, it might be worth nothing that you return a copy instead of a `const` reference due to thread-safety issues.

project3/coordinator/message_log.cpp
```diff
@@ -0,0 +1,37 @@
...
+namespace coordinator::message_log {
+
+bool MessageLog::Message::operator==(const Message &message) const {
+  return timestamp == message.timestamp && msg == message.msg &&
+         participant_id == message.participant_id;
+}
+
+size_t MessageLog::Hash::operator()(const Message &message) const {
+  // ensure uniqueness from this messages length and time stamp
+  return std::hash<int>()(message.msg.length() + message.timestamp);
```
djpetti: Do we want to include the participant ID in this hash as well? In general, it is good practice to make the hashing and equality operations correspond exactly to each-other.

project3/coordinator/message_log.h
```diff
@@ -0,0 +1,70 @@
...
+  /**
+   * Structure of messages stored by the coordinator.
+   */
+  struct Message {
+    bool operator==(const Message&) const;
+    /// The actual message.
+    std::string msg;
+
+    /// The time the message was sent.
+    double timestamp;
```
djpetti: I think storing the timestamp as a `double` here is going to be sub-optimal. For one, you're relying on this for your equality checks and floating-point values don't always compare nicely to each-other. Furthermore, C++ has the [`std::chrono::timepoint`](https://en.cppreference.com/w/cpp/chrono/time_point) class which I think suits your needs better.

project3/coordinator/message_log.h
```diff
@@ -0,0 +1,70 @@
...
+
+    /// TODO The participant who broadcast the message.
```
djpetti, line 37: Why is this a `TODO`?

project3/coordinator/message_log.h
```diff
@@ -0,0 +1,70 @@
...
+
+  /**
+   * Structure for hash function.
+   */
+  struct Hash {
+    size_t operator()(const Message& obj) const;
+  };
+  /**
+   * @brief Inserts a message into the message log.
+   * @param id The id to insert.
```
djpetti: Update parameter docs.

project3/coordinator/message_log.cpp
```diff
@@ -0,0 +1,37 @@
...
+  std::lock_guard<std::mutex> guard(mutex_);
+  coordinator_messages_.insert(msg);
+}
+void MessageLog::Clear() {
+  std::lock_guard<std::mutex> guard(mutex_);
+  coordinator_messages_.clear();
+}
+
+std::unordered_set<MessageLog::Message, MessageLog::Hash>
+MessageLog::GetMissedMessages(uint32_t reconnection_time) {
```
djpetti: Do we want this function to also automatically clear any messages that were received more than `td_` seconds ago?

project3/coordinator/message_log.cpp
```diff
@@ -0,0 +1,37 @@
...
+  std::unordered_set<MessageLog::Message, MessageLog::Hash> missed_messages;
```
djpetti, line 34: I think we need to hold the lock during this function.

project3/coordinator/message_log.h
```diff
@@ -0,0 +1,70 @@
...
+   */
+  void Insert(struct Message& msg);
```
djpetti: Should be `const struct`. Also, the `struct` specifier is unnecessary in C++.

project3/coordinator/message_log.h
```diff
@@ -0,0 +1,70 @@
...
+
+  /**
+   * @brief Clears the message log.
+   */
+  void Clear();
+
+  /**
+   * @brief Retrieves missed messages satisfying the reconnection time
+   * threshold.
+   * @param reconnection_time
```
djpetti: Finish docs.

project3/coordinator/message_log.h
```diff
@@ -0,0 +1,70 @@
...
+
+ private:
+  /// The message log.
+  std::unordered_set<Message, Hash> coordinator_messages_;
+
+  /// Mutex for implementing thread safety.
+  std::mutex mutex_;
+
+  /// The reconnection time threshold for missed messages.
+  uint8_t td_;
```
djpetti: Where does this actually get set? Shouldn't there be a constructor that sets this? Also, shouldn't it be a `double`? (Or an `std::chrono::duration` if you take my previous suggestion.)

project3/coordinator/messenger.cpp
```diff
@@ -0,0 +1,70 @@
...
+  // Pop next message from message queue.
+  MessageLog::Message msg = msg_queue_->Pop();
+
+  // Populate protobuf message.
+  proto_msg_.set_message(msg.msg);
+  proto_msg_.set_origin_id(msg.participant_id);
+
+  // Serialize the message.
+  if (!wire_protocol::Serialize(proto_msg_, &outgoing_message_buffer_)) {
+    // LOG_F(ERROR, "Failed to serialize message.");
```
djpetti: Why comment this out? This seems like a reasonable log message.

project3/coordinator/messenger.cpp
```diff
@@ -0,0 +1,70 @@
...
+  // Add this message to the message log.
+  msg_log_->Insert(msg);
+  return true;
+}
+
+bool Messenger::SendMissedMessages(uint32_t reconnection_time) {
+  for (MessageLog::Message msg :
+       msg_log_->GetMissedMessages(reconnection_time)) {
+    // Populate protobuf message.
+    proto_msg_.set_message(msg.msg);
```
djpetti: It might be worth encapsulating the whole message building, serialization, and sending process in a separate function, since it seems identical between `BroadcastMessage` and `SendMissedMessages`.

project3/coordinator/messenger.h
```diff
@@ -0,0 +1,77 @@
...
+using namespace coordinator::connected_participants;
+/**
+ * @class Handles message broadcasting to participants.
+ *
+ */
+class Messenger {
+ public:
+  /**
+   * @param msg_queue The coordinator's message queue.
+   * @param msg_log The coordinator's message log.
```
djpetti, line 29: Missing documentation for `participant`.

project3/coordinator/messenger.h
```diff
@@ -0,0 +1,77 @@
...
+#include "queue/queue.h"
+#include "wire_protocol/wire_protocol.h"
+namespace coordinator::messenger {
+using namespace coordinator::message_log;
+using namespace coordinator::connected_participants;
+/**
+ * @class Handles message broadcasting to participants.
+ *
+ */
+class Messenger {
```
djpetti, line 25: I would be tempted to say that the `Messenger` class should only be responsible for sending to a single participant, or it should only be responsible for sending to a group of them. Here you try to split the difference, having both a multicast group and a single participant that this `Messenger` is specially in-charge of. Perhaps this should be split up into two classes? Or at the very least, maybe make `SendMissedMessages` participant-agnostic, i.e. allow the receiving participant to be specified as an argument.

project3/coordinator/messenger.h
```diff
@@ -0,0 +1,77 @@
...
+
+  /**
+   * @brief Broadcasts a given message to all active participants.
+   * @note Messages sent by the messenger should be pushed in the message queue before
+   *       this function is called. Guarantees that messages will be broadcast in the order
+   *       they were received.
+   * @return true on success, false on failure
+   *
+   */
+  bool BroadcastMessage();
```
djpetti: It's a little bit of an odd design, IMO, to have this read from a queue internally. Perhaps the queue-reading should happen outside of this class, and this function should simply take the message to broadcast as an argument? That would reduce coupling too, which is always desirable.

project3/coordinator/registrar.cpp
```diff
@@ -0,0 +1,73 @@
...
+  address.sin_port = htons(port);
+
+  return address;
+}
+
+int SetUpSocket(const sockaddr_in &address, const std::string &hostname) {
+  int sock;
+
+  if ((sock = socket(AF_INET, SOCK_STREAM, 0)) < 0) {
+    perror("Socket creation error");
```
djpetti: These should be log messages instead of `perrors`.

project3/coordinator/registrar.cpp
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  return sock;
+}
+
+Registrar::Registrar(
+    std::shared_ptr<ConnectedParticipants> connected_participants)
+    : connected_participants_(connected_participants) {}
+
+void Registrar::ConnectParticipant(struct ConnectedParticipants::Participant &participant) {
```
djpetti: Again, `struct` specifier should be unnecessary in C++, and this argument should be `const`.

project3/coordinator/registrar.cpp
```diff
@@ -0,0 +1,73 @@
...
+}
+
+void Registrar::RegisterParticipant(struct ConnectedParticipants::Participant &participant) {
+  // set up socket connection for sending messages to this participant.
+  int sock = SetUpSocket(MakeAddress(participant.port), participant.hostname);
+  if (sock > 0) {
+    participant.sock_fd = sock;
+    ConnectParticipant(participant);
+  } else {
+    // Log socket creation error
```
djpetti: Why not just add the logging now?

project3/coordinator/registrar.h
```diff
@@ -0,0 +1,52 @@
...
+#include "connected_participants.h"
+#include <memory>
+
+namespace coordinator::registrar {
+using namespace connected_participants;
+/**
+ * @class Handles registration/de-registration of participants.
+ */
+class Registrar {
+ private:
```
djpetti: (minor) Usually, we put `public` before `private`, as it makes it easier for readers.

project3/coordinator/registrar.h
```diff
@@ -0,0 +1,52 @@
+/**
+ * @file Registrar to be utilized by the coordinator.
+ */
+#ifndef PROJECT3_REGISTRAR_H
+#define PROJECT3_REGISTRAR_H
+#include "connected_participants.h"
+#include <memory>
+
+namespace coordinator::registrar {
+using namespace connected_participants;
```
djpetti: Don't use `using namespace` declarations. Import each definition that you need individually, e.g. `using connected_participants::MyType`.

project3/proto/CMakeLists.txt
```diff
@@ -3,7 +3,7 @@ find_package(Protobuf REQUIRED)
 
 include_directories(${PROTOBUF_INCLUDE_DIR})
 PROTOBUF_GENERATE_CPP(PROTO_SRC PROTO_HEADER pub_sub_messages.proto)
-add_library(pub_sub_proto ${PROTO_HEADER} ${PROTO_SRC})
+add_library(pub_sub_proto ${PROTO_HEADER} ${PROTO_SRC} ../coordinator/messenger.h ../coordinator/registrar.h)
```
djpetti: Why are these header files included? I'm pretty sure this isn't correct.

djpetti (changes requested):
I think I see the overall design here, but there are definitely some changes that need to be made. I've left a mix of low-level comments, and also some high-level suggestions on how to improve your design.

djpetti:
> Do I need to make interactions with the socket connected to the participant's thread b thread-safe? It seems like there could be some concurrency issues with `BroadcastMessage()` and `SendMissedMessages()` in the `Messenger` class.

If you plan to have two threads possibly writing to the socket at the same time, then yes. Otherwise, you might get partial messages interleaved with each-other.

project3/coordinator/messenger.cpp
```diff
@@ -0,0 +1,70 @@
...
+
+  // Serialize the message.
+  if (!wire_protocol::Serialize(proto_msg_, &outgoing_message_buffer_)) {
+    // LOG_F(ERROR, "Failed to serialize message.");
+    return false;
+  }
+  // Send the message to each connected participant.
+  for (ConnectedParticipants::Participant participant :
+       connected_participants_->GetParticipants()) {
+    if (send(participant.sock_fd, outgoing_message_buffer_.data(),
```
djpetti: As you yourself pointed out, this might have concurrency issues if multiple threads are broadcasting concurrently.

project3/coordinator/messenger.h
djpetti: Actually, now I'm rethinking this suggestion a little. It seems that the way you've intended for this class to be used is to encapsulate communication with a *single* participant. To that end, the best idea is probably to make this class responsible only for sending to a single `participant`, and have a separate class that aggregates multiple `Messengers` for handling broadcasting. If you do this right, it would also solve your broadcasting concurrency issue.

djpetti (changes requested):
Re-thinking my own suggestions...

project3/coordinator/messenger.h
jake-chandler: So, bouncing off your idea here, if the the `Messenger` class is to send to one participant at a time, we could throw in a `mutex lock` in the function `SendMessage()`. Then, in the `MessengerManager` class (name in progress), to broadcast to all connected participants, we can create a `Messenger` for each connected participant, and then call the `SendMessage()' for each `Messenger`. Is this what you're referring to when you say it could solve broadcasting concurrency issues?

project3/coordinator/messenger.h
jake-chandler: Yeah, I agree with this. This also implies that any pushing/popping from the queue can all occur outside of the messenger, which means the messenger doesn't even need to store the message queue internally.

project3/coordinator/registrar.cpp
jake-chandler: Gotcha. Will fix all related issues.

jake-chandler (commented):
So, bouncing off your idea for making a separate class for multiple messengers, if the the Messenger class is to send to one participant at a time, we could throw in a mutex lock in the function SendMessage(). Then, in the MessengerManager class (name in progress), to broadcast to all connected participants, we can create a Messenger for each connected participant, and then call the SendMessage()' for each Messenger`. Is this what you're referring to when you say it could solve broadcasting concurrency issues?

project3/coordinator/messenger.h
djpetti: Yup, this is exactly what I was thinking of.

project3/coordinator/messenger.h
djpetti: Agreed. That's what I was thinking too.

project3/coordinator/CMakeLists.txt
```diff
@@ -1,2 +1,2 @@
-add_library(coordinator connected_participants.cpp message_log.cpp registrar.cpp messenger.cpp)
-target_link_libraries(coordinator pub_sub_proto wire_protocol queue)
\ No newline at end of file
+add_library(coordinator connected_participants.cpp message_log.cpp registrar.cpp messenger.cpp messenger_manager.h messenger_manager.cpp)
```
djpetti: Don't need the `.h` file here.

project3/coordinator/messenger.cpp
```diff
@@ -5,71 +5,62 @@
...
-    std::shared_ptr<ConnectedParticipants::Participant> participant,
-    std::shared_ptr<ConnectedParticipants> connected_participants)
-    : msg_queue_(std::move(msg_queue)),
-      msg_log_(std::move(msg_log)),
-      connected_participants_(std::move(connected_participants)),
-      participant_(std::move(participant)) {}
+Messenger::Messenger(std::shared_ptr<MessageLog> msg_log,
+                     ConnectedParticipants::Participant participant)
+    : msg_log_(std::move(msg_log)), participant_(std::move(participant)) {}
+bool Messenger::operator==(const Messenger& messenger) const {
```
djpetti: If you're using `shared_ptr`s in an `unordered_set`, you probably don't need this overload...

project3/coordinator/messenger.h
```diff
@@ -53,26 +48,37 @@ class Messenger {
...
+   * @brief Getter for this messenger's participant.
+   * @return This messenger's participant.
+   */
+  ConnectedParticipants::Participant GetParticipant() const;
 
  private:
-  /// The outgoing message queue.
-  std::shared_ptr < queue::Queue<MessageLog::Message>> msg_queue_;
+  /// == operator overload.
+  bool operator==(const Messenger&) const;
```
djpetti: The fact that this compiles even though this is private suggests that it's not necessary.

project3/coordinator/messenger.cpp
```diff
@@ -5,71 +5,62 @@
...
+  return participant_ == messenger.participant_;
+}
 
-bool Messenger::BroadcastMessage() {
-  // Pop next message from message queue.
-  MessageLog::Message msg = msg_queue_->Pop();
+ConnectedParticipants::Participant Messenger::GetParticipant() const {
```
djpetti: You probably want to return a `const ConnectedParticipants::Participant&` instead, right?

project3/coordinator/messenger.cpp
```diff
@@ -5,71 +5,62 @@
...
-      LOG_F(ERROR, "Failed to send message.");
-      return false;
-    }
+  return true;
+}
+bool Messenger::SendMessage(MessageLog::Message msg, bool missed_msg) {
+  SerializeMessage(msg);
+  // mutex lock to ensure thread-safe socket sending.
+  std::lock_guard<std::mutex> guard(mutex_);
+  if (send(participant_.sock_fd, outgoing_message_buffer_.data(),
```
djpetti, line 36: Since you have a `proto_msg_` that's a class member, you probably want to grab the lock at the top of this function.

project3/coordinator/messenger.h
```diff
@@ -28,23 +29,17 @@ class Messenger {
...
-   *       this function is called. Guarantees that messages will be broadcast in the order
-   *       they were received.
+   * @brief Sends a given message to this messenger's participant.
+   * @param msg The message to send.
+   * @param missed_msg if the message being sent is a missed message.
    * @return true on success, false on failure
-   *
    */
-  bool BroadcastMessage();
+  bool SendMessage(MessageLog::Message msg, bool missed_msg);
```
djpetti: Maybe the first argument should be a `const` reference? If you're actually modifying it, it really should be a pointer in accordance with the style guide. Also, perhaps a better way to do what you want is to move the missed-message-specific stuff into the `SendMissedMessages` function so you don't have to have that second argument.

project3/coordinator/messenger_manager.cpp
```diff
@@ -0,0 +1,29 @@
+/**
+ * @file Implementation of MessengerManager class.
+ */
+#include "messenger_manager.h"
+namespace coordinator {
+MessengerManager::MessengerManager(
+    std::shared_ptr<ConnectedParticipants> participants)
+    : participants_(participants) {}
+
+void MessengerManager::AddMessenger(std::shared_ptr<Messenger> messenger) {
```
djpetti: Just for thoroughness, you might want `messenger` to be a `const` reference.

project3/coordinator/messenger_manager.h
```diff
@@ -0,0 +1,46 @@
...
+ * @file Manages all active messengers.
+ */
+#ifndef CSCI6780_MESSENGER_MANAGER_H
+#define CSCI6780_MESSENGER_MANAGER_H
+#include "messenger.h"
+namespace coordinator {
+class MessengerManager {
+ public:
+
+  MessengerManager(std::shared_ptr<ConnectedParticipants> participants);
```
djpetti, line 14: (minor) Should add docs for the constructor since it has an argument.

djpetti (changes requested):
Mostly minor comments on this one. One thread-safety bug too.

project3/coordinator/messenger.h
jake-chandler: So, to keep with the thread-safety, `SendMissedMessages()` calls  `SendMessage()`, and in `SendMessage()`, we update the timestamp of the message and add it to the message log. For this reason, I think I need to keep that second argument. 
I suppose I could just create a separate function in `UpdateMessage()` that does this and have the coordinator call that after sending a message,  but I'm not sure that that is a better design choice?

djpetti: Hmm, is this really a thread-safety issue though? As far as I can tell, the only state that's getting updated there is the message timestamp, and I doubt the same message would be getting sent concurrently from multiple threads... Also, it's being added to the log, but that operation is already thread-safe. Really, the only reason I can see for doing it this way is if you needed the sending, timestamp-updating, and log-updating operations to operate as a single atomic block. Is there something I'm missing that necessitates that?

jake-chandler: Yeah, the timestamp thread-safety issue is a highly unlikely edge case. I'll change it.

project3/coordinator/messenger.cpp
```diff
@@ -29,7 +29,7 @@ bool Messenger::SerializeMessage(MessageLog::Message msg) {
   }
   return true;
 }
-bool Messenger::SendMessage(MessageLog::Message msg, bool missed_msg) {
+bool Messenger::SendMessage(const MessageLog::Message msg) {
```
djpetti: Did you mean for this to be a reference?

djpetti (approved):
Approved pending resolution of remaining comments.

PR #28, opened 2021-04-04
Add port_number to Register, Reconnect

Nickster258:
Adds the `port_number` parameter to the `Register` and `Reconnect` message.


djpetti (approved):
LGTM

PR #35, opened 2021-04-05
Add participant

### LINKED ISSUE #18: Finish Participant Code
I don't know the exact planned design for the participant, so this is sort of a catch-all ticket. It might be worth splitting it up.

Nickster258:
This adds the Participant implementation, along with all of the associated tasks necessary to run the participant.

### DIFF
--- a/project3/participant/participant.cpp
+++ b/project3/participant/participant.cpp
@@ -2,6 +2,8 @@
 
 #include <iostream>
 
+#include <loguru.hpp>
+
 namespace participant {
 Participant::Participant(const std::string& config_loc)
     : console_task_(
@@ -80,7 +82,7 @@ bool Participant::OneOff(const google::protobuf::Message& msg) {
   wire_protocol::Serialize(msg, &outgoing_buf);
   if (participant_util::SendForever(participant_fd, outgoing_buf.data(),
                                     outgoing_buf.size(), 0) < 0) {
-    perror("Failed to send request");
+    LOG_S(0) << "Failed to send request";
     return false;
   }
   close(participant_fd);
--- a/project3/participant/participant.cpp
+++ b/project3/participant/participant.cpp
@@ -1,0 +1,105 @@
+#include "participant.h"
+
+#include <iostream>
+
+namespace participant {
+Participant::Participant(const std::string& config_loc)
+    : console_task_(
+          std::make_shared<participant_tasks::ConsoleTask>("participant=> ")) {
+  LoadConfig(config_loc);
+}
+void Participant::Start() {
+  pool_.AddTask(console_task_);
+
+  while (running_) {
+    std::string input;
+    std::getline(std::cin, input);
+    input_parser_.Parse(input);
+    // Validate user input.
+    if (!input_parser_.IsValid()) {
+      console_task_->SendConsole("Invalid command!");
+      continue;
+    }
+    // If user requested to quit, exit participant peacefully
+    if (input_parser_.req_ == input_parser::InputParser::QUIT) {
+      std::cout << "Bye!" << std::endl;
+      running_ = false;
+    }
+    // Else, send the request
+    if (!OneOff(*input_parser_.CreateReq())) {
+      std::cout << "Couldn't instantiate a connection with coordinator."
+                << std::endl;
+      running_ = false;
+      continue;
+    }
+    switch (input_parser_.req_) {
+      case input_parser::InputParser::REG:
+      case input_parser::InputParser::RECON: {
+        if (connected_) {
+          console_task_->SendConsole("Already connected!");
+          break;
+        }
+        multicast_receiver_ =
+            std::make_shared<participant_tasks::MulticastReceiver>(
+                console_task_, log_location_, input_parser_.port_);
+        pool_.AddTask(multicast_receiver_);
+        connected_ = true;
+        std::string statement =
+            input_parser_.req_ == input_parser::InputParser::REG ? "Registered"
+                                                                 : "Connected";
+        console_task_->SendConsole(statement + ", listening on " +
+                                   std::to_string(input_parser_.port_));
+        break;
+      }
+      case input_parser::InputParser::DEREG:
+      case input_parser::InputParser::DISCON: {
+        if (!connected_) {
+          console_task_->SendConsole("Nothing to disconnect from!");
+          break;
+        }
+        pool_.CancelTask(multicast_receiver_);
+        connected_ = false;
+        std::string statement =
+            input_parser_.req_ == input_parser::InputParser::DEREG
+                ? "Deregistered!"
+                : "Disconnected!";
+        console_task_->SendConsole(statement);
+        break;
+      }
+      case input_parser::InputParser::MSEND:
+      default:
+        break;
+    }
+  }
+}
+
+bool Participant::OneOff(const google::protobuf::Message& msg) {
+  int participant_fd = participant_util::SetUpSocket(
+      participant_util::MakeAddress(coordinator_port_), coordinator_ip_);
+  std::vector<uint8_t> outgoing_buf;
+  wire_protocol::Serialize(msg, &outgoing_buf);
+  if (participant_util::SendForever(participant_fd, outgoing_buf.data(),
+                                    outgoing_buf.size(), 0) < 0) {
+    perror("Failed to send request");
+    return false;
+  }
+  close(participant_fd);
+  return true;
+}
+
+void Participant::LoadConfig(const std::string& config_loc) {
+  std::ifstream conf_file(config_loc);
+  if (conf_file.is_open()) {
+    std::string line;
+    std::getline(conf_file, line);
+    log_location_ = line;
+    std::getline(conf_file, line);
+    std::stringstream ss(line);
+    ss >> coordinator_ip_;
+    ss >> coordinator_port_;
+  } else {
+    std::cout << "No config or invalid config found!" << std::endl;
+    running_ = false;
+  }
+}
+}  // namespace participant
--- a/project3/participant/participant.cpp
+++ b/project3/participant/participant.cpp
@@ -5,7 +5,7 @@
 #include <loguru.hpp>
 
 namespace participant {
-Participant::Participant(const std::string& config_loc)
+Participant::Participant(const std::filesystem::path& config_loc)
     : console_task_(
           std::make_shared<participant_tasks::ConsoleTask>("participant=> ")) {
   LoadConfig(config_loc);
@@ -89,7 +89,7 @@ bool Participant::ConnectAndSend(const google::protobuf::Message& msg) {
   return true;
 }
 
-void Participant::LoadConfig(const std::string& config_loc) {
+void Participant::LoadConfig(const std::filesystem::path& config_loc) {
   std::ifstream conf_file(config_loc);
   if (conf_file.is_open()) {
     std::string line;
--- a/project3/participant/participant.cpp
+++ b/project3/participant/participant.cpp
@@ -27,15 +27,21 @@ void Participant::Start() {
       std::cout << "Bye!" << std::endl;
       running_ = false;
     }
-    // Else, send the request
-    if (!ConnectAndSend(*input_parser_.CreateReq())) {
+    // Else, open connection and send the request
+    int participant_fd = ConnectAndSend(*input_parser_.CreateReq());
+    if (participant_fd == -1) {
       std::cout << "Couldn't instantiate a connection with coordinator."
                 << std::endl;
       running_ = false;
       continue;
     }
     switch (input_parser_.req_) {
-      case input_parser::InputParser::REG:
+      case input_parser::InputParser::REG: {
+        WaitForMessage(participant_fd);
+        pub_sub_messages::RegistrationResponse response;
+        parser_.GetMessage(&response);
+        input_parser_.participant_id_ = response.participant_id();
+      }
       case input_parser::InputParser::RECON: {
         if (connected_) {
           console_task_->SendConsole("Already connected!");
@@ -72,10 +78,29 @@ void Participant::Start() {
       default:
         break;
     }
+    // Close participant connection
+    close(participant_fd);
   }
 }
 
-bool Participant::ConnectAndSend(const google::protobuf::Message& msg) {
+void Participant::WaitForMessage(int fd_) {
+  parser_.ResetParser();
+  std::vector<uint8_t> outgoing_msg_buf{};
+  int buf_size = 4096;
+  while (!parser_.HasCompleteMessage()) {
+    outgoing_msg_buf.resize(buf_size);
+    const auto bytes_read =
+        participant_util::ReceiveForever(fd_, outgoing_msg_buf.data(), buf_size, 0);
+    if (bytes_read <= 0) {
+      break;
+    }
+    outgoing_msg_buf.resize(bytes_read);
+    parser_.AddNewData(outgoing_msg_buf);
+  }
+  outgoing_msg_buf.clear();
+}
+
+int Participant::ConnectAndSend(const google::protobuf::Message& msg) {
   int participant_fd = participant_util::SetUpSocket(
       participant_util::MakeAddress(coordinator_port_), coordinator_ip_);
   std::vector<uint8_t> outgoing_buf;
@@ -83,10 +108,9 @@ bool Participant::ConnectAndSend(const google::protobuf::Message& msg) {
   if (participant_util::SendForever(participant_fd, outgoing_buf.data(),
                                     outgoing_buf.size(), 0) < 0) {
     LOG_S(0) << "Failed to send request";
-    return false;
+    return -1;
   }
-  close(participant_fd);
-  return true;
+  return participant_fd;
 }
 
 void Participant::LoadConfig(const std::filesystem::path& config_loc) {
--- a/project3/participant/participant.cpp
+++ b/project3/participant/participant.cpp
@@ -28,7 +28,7 @@ void Participant::Start() {
       running_ = false;
     }
     // Else, send the request
-    if (!OneOff(*input_parser_.CreateReq())) {
+    if (!ConnectAndSend(*input_parser_.CreateReq())) {
       std::cout << "Couldn't instantiate a connection with coordinator."
                 << std::endl;
       running_ = false;
@@ -75,7 +75,7 @@ void Participant::Start() {
   }
 }
 
-bool Participant::OneOff(const google::protobuf::Message& msg) {
+bool Participant::ConnectAndSend(const google::protobuf::Message& msg) {
   int participant_fd = participant_util::SetUpSocket(
       participant_util::MakeAddress(coordinator_port_), coordinator_ip_);
   std::vector<uint8_t> outgoing_buf;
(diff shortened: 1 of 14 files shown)


project3/participant/participant.h
```diff
@@ -0,0 +1,62 @@
...
+  /**
+   * @param msg the message to send to the coordinator
+   * @return successfully sent
+   */
+  bool OneOff(const google::protobuf::Message& msg);
+
+  /// Pool and tasks
+  thread_pool::ThreadPool pool_;
+  std::shared_ptr<participant_tasks::ConsoleTask> console_task_;
+  std::shared_ptr<participant_tasks::MulticastReceiver> multicast_receiver_;
```
Nickster258, line 54: Note: this is one example of a design compromise I made to satisfy project 3 requirements. If this has to be expanded on for a later project, I expect to refactor this accordingly.

project3/participant/tasks/multicast_receiver_task.cpp
```diff
@@ -0,0 +1,46 @@
...
+namespace participant_tasks {
+MulticastReceiver::MulticastReceiver(
+    const std::shared_ptr<ConsoleTask>& console_task, std::string log_location,
+    int port)
+    : port_(port), log_location_(log_location), console_task_(console_task) {}
+
+thread_pool::Task::Status MulticastReceiver::SetUp() {
+  struct sockaddr_in address = participant_util::MakeAddress(port_);
+  messenger_fd_ = participant_util::SetUpListenerSocket(address);
+  loguru::add_file(log_location_.c_str(), loguru::FileMode::Truncate,
```
Nickster258, line 14: I wasn't sure what log level to log here. I want this log to **only** log messages received by this task.

project3/participant/participant.cpp
```diff
@@ -0,0 +1,105 @@
...
+  if (participant_util::SendForever(participant_fd, outgoing_buf.data(),
+                                    outgoing_buf.size(), 0) < 0) {
+    perror("Failed to send request");
+    return false;
+  }
+  close(participant_fd);
+  return true;
+}
+
+void Participant::LoadConfig(const std::string& config_loc) {
```
Nickster258: This assumes the log file is formatted as follows:

```
/path/to/log.txt
hostname port
```

Nickster258:
Also, I wasn't sure what was the best way to include an "example" config file. I was trying to think of how one can package resources in a Jar, but that is obviously not applicable here.

project3/participant/participant.cpp
```diff
@@ -0,0 +1,105 @@
+#include "participant.h"
+
+#include <iostream>
+
+namespace participant {
+Participant::Participant(const std::string& config_loc)
+    : console_task_(
+          std::make_shared<participant_tasks::ConsoleTask>("participant=> ")) {
```
djpetti, line 10: (minor) Might be worth making this prompt a constant.

project3/participant/participant.h
```diff
@@ -0,0 +1,62 @@
...
+
+#include "../../common/thread_pool/task.h"
+#include "../../common/thread_pool/thread_pool.h"
+#include "input_parser/input_parser.h"
+#include "pub_sub_messages.pb.h"
+#include "tasks/console_task.h"
+#include "tasks/multicast_receiver_task.h"
+
+namespace participant {
+class Participant {
```
djpetti, line 22: (minor) Class should have top-level docs.

project3/participant/participant.h
```diff
@@ -0,0 +1,62 @@
...
+ public:
+  /**
+   * @param config_loc the location of the config file
+   */
+  explicit Participant(const std::string& config_loc);
```
djpetti: (minor) Prefer `std::filesystem::path` for representing paths.

project3/participant/participant.cpp
```diff
@@ -0,0 +1,105 @@
...
+}
+
+bool Participant::OneOff(const google::protobuf::Message& msg) {
+  int participant_fd = participant_util::SetUpSocket(
+      participant_util::MakeAddress(coordinator_port_), coordinator_ip_);
+  std::vector<uint8_t> outgoing_buf;
+  wire_protocol::Serialize(msg, &outgoing_buf);
+  if (participant_util::SendForever(participant_fd, outgoing_buf.data(),
+                                    outgoing_buf.size(), 0) < 0) {
+    perror("Failed to send request");
```
djpetti: This should be a log instead of `perror`.

project3/participant/participant.h
```diff
@@ -0,0 +1,62 @@
...
+  /**
+   * @param config_loc the config file location
+   */
+  void LoadConfig(const std::string& config_loc);
+
+  /**
+   * @param msg the message to send to the coordinator
+   * @return successfully sent
+   */
+  bool OneOff(const google::protobuf::Message& msg);
```
djpetti: (minor) I'm not a huge fan of the name of this method. Perhaps this should be called `ConnectAndSend` or something like that?

project3/participant/participant.cpp
djpetti: The path to the log file is a custom extension on our part, right? If so, I would suggest that you make it so that it still works without a log file specified, otherwise I foresee more emails from the TA in the near future.

project3/participant/participant_main.cpp
```diff
@@ -0,0 +1,32 @@
+/**
+ * @brief Entry point for the Multicast Participant.
```
djpetti: (minor) `@file`

project3/participant/participant.h
```diff
@@ -0,0 +1,62 @@
...
+
+  /// Pool and tasks
+  thread_pool::ThreadPool pool_;
+  std::shared_ptr<participant_tasks::ConsoleTask> console_task_;
+  std::shared_ptr<participant_tasks::MulticastReceiver> multicast_receiver_;
+
+  /// Config options
+  std::string log_location_;
```
djpetti, line 57: It looks like the log path gets set as a CLI argument also... Is the version from the config file actually used?

project3/participant/participant_util.cpp
```diff
@@ -0,0 +1,113 @@
...
+  address.sin_port = htons(port);
+
+  return address;
+}
+
+int SetUpListenerSocket(const struct sockaddr_in &address) {
+  // Open a TCP socket.
+  const int server_fd = socket(AF_INET, SOCK_STREAM, 0);
+  if (server_fd == 0) {
+    perror("Failed to create server socket");
```
djpetti: These should be logs instead of `perror`. I feel like I see this code over-and-over again. Maybe later we should break this out into `common`.

project3/participant/participant_util.cpp
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  return server_fd;
+}
+
+int SetUpSocket(const sockaddr_in &address, const std::string &hostname) {
+  int sock = 0;
+
+  if ((sock = socket(AF_INET, SOCK_STREAM, 0)) < 0) {
+    perror("Socket creation error");
```
djpetti: Same comment about logging vs. `perror`.

project3/participant/participant_util.cpp
```diff
@@ -0,0 +1,113 @@
...
+
+  if (connect(sock, (struct sockaddr *)&address, sizeof(address)) < 0) {
+    perror("Connection Failed");
+    return -1;
+  }
+
+  return sock;
+}
+
+int ReceiveForever(int socket, void *buffer, size_t length, int flags) {
```
djpetti, line 74: I recall writing these utilities originally for project 2. I assume they were copied/pasted here. At some point, I should probably move them into `common` if they're going to be used for multiple projects.

Also, these functions are superfluous if you disable timeouts on the socket.

project3/participant/participant_util.h
```diff
@@ -0,0 +1,58 @@
...
+#include <fstream>
+
+#include <pub_sub_messages.pb.h>
+#include "wire_protocol/wire_protocol.h"
+
+namespace participant_util {
+
+/**
+ * @param port the port of the address
+ * @return
```
djpetti: (minor) Finish docs

project3/participant/participant_util.h
```diff
@@ -0,0 +1,58 @@
...
+ */
+struct sockaddr_in MakeAddress(uint16_t port);
+
+/**
+ * @param address the address to connect to
+ * @param hostname the host to connect to
+ * @return the socket file descriptor
```
djpetti, line 24: (minor) For this and the function below, it might be worth mentioning that it returns -1 on failure.

project3/participant/participant_util.cpp
```diff
@@ -0,0 +1,113 @@
...
+                (struct sockaddr *)&address.sin_addr) <= 0) {
+    perror("Invalid Address or Address not supported");
+    return -1;
+  }
+
+  // Set a timeout.
+  struct timeval timeout {};
+  timeout.tv_sec = kSocketTimeout;
+  timeout.tv_usec = 0;
+  setsockopt(sock, SOL_SOCKET, SO_RCVTIMEO, (const char *)&timeout,
```
djpetti: Do we want to use timeouts on the sockets for this project? IIRC, the timeouts were added for project 2 to deal with a specific issue related to cancelled file transfers. It seems like this is not necessary for project 3.

project3/participant/tasks/console_task.cpp
```diff
@@ -0,0 +1,38 @@
...
+  }
+  // Clear console output buffer
+  console_msg_buf_.erase(console_msg_buf_.begin(), console_msg_buf_.end());
+  // Display prompt for end-user
+  std::cout << prompt_ << std::flush;
+  return thread_pool::Task::Status::RUNNING;
+}
+
+void ConsoleTask::SendConsole(const std::string& message) {
+  console_msg_buf_.push_back(message);
```
djpetti: This is not thread-safe. I suggest you use a thread-safe queue here instead of a vector.

project3/participant/tasks/console_task.h
```diff
@@ -0,0 +1,44 @@
...
+
+#include <cstdint>
+#include <iostream>
+#include <string>
+#include <vector>
+
+#include "thread_pool/task.h"
+
+namespace participant_tasks {
+class ConsoleTask : public thread_pool::Task {
```
djpetti, line 22: (minor) Should have top-level docs for class

project3/participant/tasks/multicast_receiver_task.cpp
djpetti: I think level 0 here is reasonable. I'm not sure I see a huge advantage to having a separate log file for this task, though. Loguru is inherently thread-safe, so you don't have to worry about that.

project3/participant/tasks/multicast_receiver_task.cpp
```diff
@@ -0,0 +1,46 @@
...
+        messenger_fd_, incoming_msg_buf_.data(), kBufferSize, 0);
+    if (bytes_read <= 0) {
+      break;
+    }
+    incoming_msg_buf_.resize(bytes_read);
+    parser_.AddNewData(incoming_msg_buf_);
+  }
+  // empty buffer
+  incoming_msg_buf_.clear();
+  if (parser_.HasCompleteMessage()) {
```
djpetti, line 21: Isn't this condition always going to be true by the time the above loop exits? From a design perspective, it's probably better to not reset the parser on every call to this function, and just have it run one iteration of the receiving loop on each `RunAtomic` call. It's simpler code and it fits more with the `Task` design philosophy.

project3/participant/tasks/multicast_receiver_task.h
```diff
@@ -0,0 +1,52 @@
...
+#include <cstdint>
+#include <memory>
+#include <string>
+
+#include "../participant_util.h"
+#include "console_task.h"
+#include "thread_pool/task.h"
+
+namespace participant_tasks {
+class MulticastReceiver : public thread_pool::Task {
```
djpetti, line 26: (minor) Class should have top-level docs.

project3/participant/tasks/multicast_receiver_task.h
```diff
@@ -0,0 +1,52 @@
...
+  std::shared_ptr<ConsoleTask> console_task_;
+
+  /// buffer size for client.
+  static constexpr size_t kBufferSize = 4096;
+
+  /// buffer that stores serialized data to be received from the coordinator
+  std::vector<uint8_t> incoming_msg_buf_{};
+
+  /// parser for handling messages
+  wire_protocol::MessageParser<google::protobuf::Message> parser_;
```
djpetti: Shouldn't the template parameter here be `ForwardMulticast`?

djpetti (changes requested):
Mostly minor comments... A few miscellaneous bugs

project3/participant/tasks/multicast_receiver_task.cpp
Nickster258: It was more to satisfy part of the project description. Namely, the part where participants are supposed to log messages sent over through that group while they are registered.

project3/participant/participant.h
Nickster258: Ah, never even heard of that, I'll look into it.

project3/participant/participant.h
Nickster258: Yeah, the method name was named after my brainstorming of the concept. I think your suggestion is a better name.

project3/participant/tasks/multicast_receiver_task.h
Nickster258: You're right, it can definitely be `ForwardMulticast`. I think I originally had other ideas for this parser but never adjusted it later.

project3/participant/participant_util.cpp
Nickster258: > I assume they were copied/pasted here.

Basically. Though, I shied away from introducing any new `common` library for the sake of time.

project3/participant/tasks/multicast_receiver_task.cpp
djpetti: Ah, that's my fault for not reading the description. After reading it, I think that this is probably a decent way to go. However, as it stands, I think this will also capture log messages from every other thread. We might have to log messages manually to mitigate that.

project3/participant/participant.cpp
djpetti: Nvm, I see that this is actually part of the project.

project3/participant/participant.h
Nickster258: The only CLI parameter is the config parameter. It then populates log_location_ from said config. https://github.com/djpetti/csci6780/blob/feature/participant/project3/participant/participant.cpp#L95

project3/participant/participant.h
djpetti: We used it in project 1 and 2 in the `FileHandler` class.

project3/participant/tasks/multicast_receiver_task.cpp
Nickster258: I should be more specific. By "log messages" I meant multicast messages sent through the group.

project3/participant/participant.h
djpetti: Right, I meant to delete this comment... Yeah, this is fine.

project3/participant/participant_util.cpp
djpetti: Right. The thing about `common` was more a note to myself about something that I should do in the future. I don't expect you to do that now.

djpetti:
> Also, I wasn't sure what was the best way to include an "example" config file. I was trying to think of how one can package resources in a Jar, but that is obviously not applicable here.

In the past, I've had an `examples` directory in the `git` repo which can include example config files.

jake-chandler (approved):

project3/participant/participant_util.cpp
```diff
@@ -25,25 +24,25 @@ int SetUpListenerSocket(const struct sockaddr_in &address) {
   // Open a TCP socket.
   const int server_fd = socket(AF_INET, SOCK_STREAM, 0);
   if (server_fd == 0) {
-    perror("Failed to create server socket");
+    LOG_S(FATAL) << "Failed to create server socket";
```
djpetti: You might want to make these ERROR instead of FATAL. I believe the latter will abort the program immediately.

project3/participant/tasks/console_task.h
```diff
@@ -10,16 +10,21 @@
 #include <iostream>
 #include <string>
 #include <vector>
+#include "../../../common/queue/queue.h"
```
djpetti, line 13: The way the include directories are set up, you should be able to just do `#include common/queue/queue.h`.

project3/participant/tasks/console_task.h
```diff
@@ -36,7 +41,7 @@ class ConsoleTask : public thread_pool::Task {
    */
   void ClearLine();
 
-  std::vector<std::string> console_msg_buf_{};
+  queue::Queue<std::string> console_msg_buf_{};
```
djpetti: (minor) Might want to call this `console_message_queue_` since it's not really a buffer anymore.

project3/participant/tasks/multicast_receiver_task.cpp
```diff
@@ -17,26 +17,28 @@ thread_pool::Task::Status MulticastReceiver::SetUp() {
...
   if (parser_.HasCompleteMessage()) {
     pub_sub_messages::ForwardMulticast msg;
     parser_.GetMessage(&msg);
     std::string to_out =
         "[" + std::to_string(msg.origin_id()) + "] " + msg.message();
     LOG_S(0) << to_out;
     console_task_->SendConsole(to_out);
+    parser_.ResetParser();
+    incoming_msg_buf_.clear();
+  // Else, attempt to parse_ anything new
```
djpetti: "attempt to parse"?

project3/participant/tasks/multicast_receiver_task.cpp
```diff
@@ -17,26 +17,28 @@ thread_pool::Task::Status MulticastReceiver::SetUp() {
...
-  incoming_msg_buf_.clear();
+  // If parser_ has a completed message ready
   if (parser_.HasCompleteMessage()) {
     pub_sub_messages::ForwardMulticast msg;
     parser_.GetMessage(&msg);
     std::string to_out =
         "[" + std::to_string(msg.origin_id()) + "] " + msg.message();
     LOG_S(0) << to_out;
     console_task_->SendConsole(to_out);
+    parser_.ResetParser();
```
djpetti: Technically, `GetMessage` does this internally, so this line should be superfluous.

project3/participant/tasks/multicast_receiver_task.cpp
```diff
@@ -17,26 +17,28 @@ thread_pool::Task::Status MulticastReceiver::SetUp() {
...
+    incoming_msg_buf_.clear();
+  // Else, attempt to parse_ anything new
+  } else {
+    incoming_msg_buf_.resize(kBufferSize);
+    const auto bytes_read = participant_util::ReceiveForever(
+        messenger_fd_, incoming_msg_buf_.data(), kBufferSize, 0);
+    if (bytes_read < 0) {
+      return thread_pool::Task::Status::FAILED;
+    } else if (bytes_read == 0) {
+      return thread_pool::Task::Status::RUNNING;
```
djpetti: This means the socket was closed. I think you want to return `Task::Status::DONE` in this case. Might be worth logging something also...

djpetti (changes requested):
Mostly minor comments. Should be quick fixes.

djpetti (approved):

djpetti (approved):

PR #37, opened 2021-04-05
Feature/coordinator

jake-chandler:
Rough draft of the Coordinator should be ready for testing with the `Participant`. Note that I've refactored the logic and protobuf messages such that the Participant informs the Coordinator of it's ID. You'll see that this design takes inspiration from `Agent` from p2.

Also, I believe I'm using the `Timestamp` and `Duration` correctly, but there might be something I missed.


project3/coordinator/CMakeLists.txt
```diff
@@ -1,2 +1,8 @@
...
-target_link_libraries(coordinator pub_sub_proto wire_protocol loguru)
\ No newline at end of file
+add_subdirectory(coordinator_resources)
+add_library(coordinator coordinator.cpp coordinator.cpp
+        coordinator_task.cpp coordinator_main.cpp
+        coordinator_driver.cpp)
+target_link_libraries(coordinator coordinator_resources pub_sub_proto wire_protocol loguru thread_pool)
+
+add_executable(coordinator_main coordinator_main.cpp)
+target_link_libraries(coordinator_main coordinator_resources coordinator loguru)
```
djpetti: I'm surprised you have to link with `coordinator_resources` here.

project3/coordinator/Config.txt
```diff
@@ -0,0 +1,2 @@
+40
```
djpetti, line 1: um

project3/coordinator/coordinator.cpp
```diff
@@ -0,0 +1,157 @@
...
+  return ClientState::ACTIVE;
+}
+
+Coordinator::ClientState Coordinator::DispatchMessage(
+    const pub_sub_messages::CoordinatorMessage &message) {
+  if (message.has_register_()) {
+    LOG_F(INFO, "Handling a register command from Participant %i.",participant_.id);
+    return HandleRequest(message.register_());
+  } else if (message.has_deregister()) {
+    LOG_F(INFO, "Handling a reregister command from Participant %i.",participant_.id);
```
djpetti: Do you mean `deregister` in the log message?

project3/coordinator/coordinator.cpp
```diff
@@ -0,0 +1,157 @@
...
+}
+
+Coordinator::ClientState Coordinator::HandleRequest(const pub_sub_messages::SendMulticast &request) {
+  MessageLog::Message message;
+  // create and broadcast the message.
+  Timestamp timestamp;
+  message.timestamp = timestamp;
+  message.msg = request.message();
+  message.participant_id = participant_.id;
+  msg_queue_->Push(message);
```
djpetti, line 193: (minor) Might be worth adding a comment here specifying that the queue push-pop is to ensure that messages from all threads get sent in the same order that they were received.

project3/coordinator/coordinator.h
```diff
@@ -0,0 +1,88 @@
...
+ */
+
+#ifndef CSCI6780_COORDINATOR_H
+#define CSCI6780_COORDINATOR_H
+#include "coordinator_resources/messenger.h"
+#include "coordinator_resources/messenger_manager.h"
+#include "coordinator_resources/registrar.h"
+
+namespace coordinator {
+class Coordinator {
```
djpetti, line 15: (minor) Class should have top-level docs.

project3/coordinator/coordinator.h
```diff
@@ -0,0 +1,88 @@
...
+ public:
+  Coordinator(int participant_fd, std::string hostname,
```
djpetti, line 17: (minor) Docs are needed for this constructor.

project3/coordinator/coordinator.h
```diff
@@ -0,0 +1,88 @@
...
+                           std::shared_ptr<MessengerManager> msg_mgr,
+                           std::shared_ptr<Registrar> registrar,
+              std::shared_ptr<queue::Queue<MessageLog::Message>> msg_queue,
+              std::shared_ptr<MessageLog> msg_log);
+
+  bool Coordinate();
```
djpetti, line 23: (minor) Missing docs

project3/coordinator/coordinator_driver.cpp
```diff
@@ -0,0 +1,89 @@
...
+
+int CoordinatorDriver::SetUpSocket(const struct sockaddr_in &address) {
+  // Open a TCP socket.
+  const int server_fd = socket(AF_INET, SOCK_STREAM, 0);
+  if (server_fd == 0) {
+    LOG_F(ERROR, "Failed to create server socket");
+    return -1;
+  }
+
+  // Allow the server to re-bind to this port if it was restarted quickly.
```
djpetti, line 33: nice

project3/coordinator/coordinator_driver.cpp
```diff
@@ -0,0 +1,89 @@
...
+  const int option = 1;
+  if (setsockopt(server_fd, SOL_SOCKET, SO_REUSEADDR | SO_REUSEPORT, &option,
+                 sizeof(option))) {
+    LOG_F(ERROR, "Failed to set socket options");
```
djpetti, line 37: Maybe this should be a WARNING then?

project3/coordinator/coordinator_driver.cpp
```diff
@@ -0,0 +1,89 @@
...
+  const auto kAddress = MakeAddress(port);
+  return SetUpSocket(kAddress);
+}
+
+[[noreturn]] void CoordinatorDriver::Start(uint16_t port, std::chrono::steady_clock::duration threshold) {
+  LOG_F(INFO, "Initializing socket...");
+  // create socket.
+  int server_fd = CreateSocket(port);
+  // initialize data structures for coordinator tasks.
+  participants_ = std::make_shared<coordinator::ConnectedParticipants>();
```
djpetti, line 65: I think these should be set in the constructor. I suggest that you do also use DI here, because it makes testing easier.

project3/coordinator/coordinator_driver.h
```diff
@@ -0,0 +1,58 @@
...
+#include "coordinator_task.h"
+#include "queue/queue.h"
+#include "thread_pool/task.h"
+#include "thread_pool/thread_pool.h"
+
+namespace coordinator {
+using coordinator::MessengerManager;
+using coordinator::Registrar;
+using coordinator::MessageLog;
+using Timestamp = std::chrono::steady_clock::time_point;
```
djpetti, line 20: Don't redefine these. Use the ones that are defined in `MessageLog`.

project3/coordinator/coordinator_driver.h
```diff
@@ -0,0 +1,58 @@
...
+  uint8_t static constexpr kMaxQueueSize_ = 5;
+
+  /// The thread pool.
+  thread_pool::ThreadPool pool_;
+
+  /**
+   * Helper functions for creating sockets.
+   *
+   */
+  static struct sockaddr_in MakeAddress(int port);
```
djpetti, line 52: (minor) Prefer functions in an anonymous namespace instead of static methods. Also, these functions should really have individual docs.

project3/coordinator/coordinator_main.cpp
```diff
@@ -0,0 +1,36 @@
...
+    std::string line;
+    std::getline(conf_file, line);
+    port = std::stoi(line);
+    std::getline(conf_file, line);
+    int time = std::stoi(line);
+
+    auto driver = std::make_shared<coordinator::CoordinatorDriver>();
+    driver->Start(port, coordinator::Duration ((long) time));
+  } else {
+    LOG_F(INFO,"Config file not found");
```
djpetti, line 26: This is an error that should really result in the whole program terminating with a non-zero exit status.

project3/coordinator/coordinator_main.cpp
```diff
@@ -0,0 +1,36 @@
...
+  }
+}
+int main(int argc, char**argv) {
+  if (argc != 2) {
+    LOG_F(ERROR, "Incorrect number of program arguments.");
+    return 0;
```
djpetti, line 32: return 1 here to signify an error.

project3/coordinator/coordinator_resources/message_log.h
```diff
@@ -49,7 +49,7 @@ class MessageLog {
    * @param time_threshold Any messages older than this will be dropped
    *    when sending messages to reconnecting participants.
    */
-  explicit MessageLog(const Duration& time_threshold);
```
djpetti, line 52: Why is this not longer `explicit`?

project3/coordinator/coordinator_resources/messenger.cpp
```diff
@@ -11,7 +11,7 @@
 namespace coordinator {
 
 Messenger::Messenger(std::shared_ptr<MessageLog> msg_log,
-                     ConnectedParticipants::Participant participant)
+                     ConnectedParticipants::Participant& participant)
```
djpetti, line 14: If it's a reference, it should be `const`. If you need to modify it, it should be a pointer.

project3/coordinator/coordinator_resources/messenger.cpp
```diff
@@ -29,7 +29,7 @@ bool Messenger::SerializeMessage(MessageLog::Message msg) {
   }
   return true;
 }
-bool Messenger::SendMessage(MessageLog::Message msg, bool missed_msg) {
+bool Messenger::SendMessage(MessageLog::Message msg) {
```
djpetti: Should `msg` be a `const` reference?

project3/coordinator/coordinator_resources/messenger.cpp
```diff
@@ -38,26 +38,27 @@ bool Messenger::SendMessage(MessageLog::Message msg, bool missed_msg) {
...
-    const auto timestamp = std::chrono::steady_clock::now();
-    msg.timestamp = timestamp;
-    // Add this message to the message log.
-    msg_log_->Insert(msg);
-  }
-
   return true;
 }
-
+void Messenger::LogMessage(MessageLog::Message& msg) {
```
djpetti: `msg` should be a pointer since it's being modified.

project3/coordinator/coordinator_resources/messenger_manager.h
```diff
@@ -20,19 +20,19 @@ class MessengerManager {
    * @return true on success, false on failure
    *
    */
-  bool BroadcastMessage(MessageLog::Message& message);
+  bool BroadcastMessage(MessageLog::Message message);
```
djpetti: `message` should probably be a `const` reference.

project3/coordinator/coordinator_task.cpp
```diff
@@ -0,0 +1,38 @@
...
+  std::string thread_name = "Coordinator Thread for client #" + std::to_string(participant_fd_);
+  loguru::set_thread_name(thread_name.c_str());
+  coordinator_ = std::make_unique<Coordinator>(
+      participant_fd_, hostname_, msg_mgr_, registrar_, msg_queue_, msg_log_);
+
+  return Status::RUNNING;
+}
+
+thread_pool::Task::Status CoordinatorTask::RunAtomic() {
+  if (coordinator_->Coordinate()) {
```
djpetti, line 33: (minor) This can be done more compactly with the ternary operator if you wish.

project3/coordinator/coordinator_task.h
```diff
@@ -0,0 +1,57 @@
...
+#include "thread_pool/task.h"
+#include "thread_pool/thread_pool.h"
+#include "coordinator.h"
+namespace coordinator {
+class CoordinatorTask : public thread_pool::Task {
+ public:
+  /**
+   *
+   * @param participant_fd The socket
+   * @param hostname
```
djpetti: Finish docs

djpetti (changes requested):
Lots of small things here and there

project3/coordinator/Config.txt
jake-chandler: Bad port # ?

djpetti: Sorry, I should have been more specific. On Linux, ports <1024 are reserved for the OS and core services.

project3/coordinator/coordinator.cpp
```diff
@@ -18,30 +19,46 @@ Coordinator::Coordinator(
...
+      msg_log_(std::move(msg_log)),
       client_fd_(participant_fd),
-      hostname_(hostname),
-      msg_mgr_(msg_mgr),
-      registrar_(registrar){}
+      hostname_(std::move(hostname)),
+      messenger_mgr_(std::move(msg_mgr)),
+      registrar_(std::move(registrar)) {}
 
 bool Coordinator::Coordinate() {
```
djpetti, line 29: Yeah, I like this a little better. Since you're running this as a task, I think this design is a better fit.

project3/coordinator/coordinator.cpp
```diff
@@ -18,30 +19,46 @@ Coordinator::Coordinator(
...
 }
 
+void Coordinator::DetermineParticipant() {
+  // determine the participant.
+  for (auto participant :
+      registrar_->GetConnectedParticipants()->GetParticipants()) {
+    if (participant.hostname == hostname_) {
+      participant_ = participant;
+    }
+  }
```
djpetti, line 52: It would make me feel better to have a `CHECK` here making sure you didn't go through the loop without modifying anything.

project3/coordinator/coordinator.cpp
```diff
@@ -82,19 +99,30 @@ Coordinator::ClientState Coordinator::ReadNextMessage(
 Coordinator::ClientState Coordinator::DispatchMessage(
     const pub_sub_messages::CoordinatorMessage &message) {
   if (message.has_register_()) {
-    LOG_F(INFO, "Handling a register command from Participant %i.",participant_.id);
+    LOG_F(INFO, "Handling a register command from Participant %i.",
+          participant_.id);
     return HandleRequest(message.register_());
   } else if (message.has_deregister()) {
-    LOG_F(INFO, "Handling a deregister command from Participant %i.",participant_.id);
+    DetermineParticipant();
```
djpetti: I'm of the opinion that perhaps these should be called in the actual `HandleRequest` overload where they are needed. Also, since they (theoretically) change with every request, is there any reason anymore to store the participant and `Messenger` in member variables? Maybe you should just make these pure functions that return the relevant instance.

project3/coordinator/coordinator_resources/messenger.h
```diff
@@ -38,7 +38,7 @@ class Messenger {
    * @param msg The message to send.
    * @return true on success, false on failure
    */
-  bool SendMessage(MessageLog::Message msg);
+  bool SendMessage(const MessageLog::Message msg);
```
djpetti: I think you mean for `msg` to be a reference.

project3/coordinator/coordinator_resources/messenger_manager.cpp
```diff
@@ -33,7 +37,7 @@ bool MessengerManager::BroadcastMessage(MessageLog::Message msg) {
       // only log message once.
       // message will have timestamp of the first message sent.
       if (!is_logged) {
-        messenger->LogMessage(msg);
+        messenger->LogMessage(const_cast<MessageLog::Message*>(&msg));
```
djpetti: `const_cast` is really not something that should be used lightly. How about making `BroadcastMessage` take a pointer to `msg` instead? The way it is now, it tricks the caller into thinking that `msg` won't be modified, when in reality its timestamp will be updated.

project3/coordinator/coordinator_resources/messenger_manager.h
```diff
@@ -33,6 +33,13 @@ class MessengerManager {
    * @param messenger
    */
   void DeleteMessenger(const std::shared_ptr<Messenger>& messenger);
+
+  /**
+   * @brief Getter for the set of Messengers.
+   * @return The set of messengers.
+   * @note for read only.
+   */
+  const std::unordered_set<std::shared_ptr<Messenger>> GetMessengers();
```
djpetti: I assume this is supposed to return a reference?

djpetti (changes requested):
We already talked about including participant IDs in the messages, so I didn't comment on that here. I do have some unrelated comments, though.

project3/coordinator/Config.txt
jake-chandler: No worries, it made me chuckle.

project3/coordinator/coordinator.cpp
jake-chandler: I did this because mainly because CLion was mentioning that using `std::move` was "Clang Tidy".

project3/coordinator/coordinator_resources/messenger_manager.h
jake-chandler: I'm not sure. I intended to have this return a copy of the `Messenger` set housed in the `MessengerManager`. From my understanding, the nature of `shared_ptr` would mean that returning a copy of this `unordered_set` would still allow the caller to access/modify values that the `shared_ptr` points to, since copies of the `shared_ptr` would point to the same thing. This thinking is what made me decide to add the `const`

project3/coordinator/coordinator.cpp
jake-chandler: Agreed.

project3/coordinator/coordinator.cpp
jake-chandler: I think this should be resolved since I had these functions return their respective types to remove the `message_` and `participant_` members. So I'm not really modifying anything in these loops.

project3/coordinator/coordinator_resources/messenger.cpp
```diff
@@ -18,21 +18,21 @@ const ConnectedParticipants::Participant& Messenger::GetParticipant() const {
...
+  proto_msg_.set_origin_id(msg->participant_id);
 
   // Serialize
   if (!wire_protocol::Serialize(proto_msg_, &outgoing_message_buffer_)) {
     return false;
   }
   return true;
 }
-bool Messenger::SendMessage(const MessageLog::Message msg) {
+bool Messenger::SendMessage(MessageLog::Message &msg) {
```
djpetti: Okay, this should be a `const` reference though, right?

project3/coordinator/coordinator_resources/messenger.cpp
```diff
@@ -50,7 +50,7 @@ void Messenger::LogMessage(MessageLog::Message *msg) {
 bool Messenger::SendMissedMessages(
     const MessageLog::Timestamp& reconnection_time) {
   int msg_count = 0;
-  for (const MessageLog::Message& msg :
+  for (MessageLog::Message msg :
```
djpetti, line 53: I think one you fix the previous comment this can be a `const` reference again also.

djpetti (approved):
Approved pending minor revisions. Will have to look into the failing build though.

project3/coordinator/coordinator_resources/messenger_manager.h
djpetti: Hmm, yeah, I guess there is the thread-safety issue here. However, this is still a rep exposure, even with the `const`, because that only applies to the `shared_ptr` itself, not the thing it's actually pointing to. I think what you want to do instead is return a normal (non-pointer) `std::unordered_set` that is a copy of the internal set of messengers.

PR #39, opened 2021-04-06
Add file logging to multicast receiver

Nickster258:
Add file logging to multicast receiver

### DIFF
--- a/project3/participant/tasks/multicast_receiver_task.cpp
+++ b/project3/participant/tasks/multicast_receiver_task.cpp
@@ -4,15 +4,13 @@
 
 namespace participant_tasks {
 MulticastReceiver::MulticastReceiver(
-    const std::shared_ptr<ConsoleTask>& console_task, std::string log_location,
-    int port)
+    const std::shared_ptr<ConsoleTask>& console_task,
+    std::filesystem::path log_location, int port)
     : port_(port), log_location_(log_location), console_task_(console_task) {}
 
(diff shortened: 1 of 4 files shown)


project3/proto/pub_sub_messages.proto
```diff
@@ -8,6 +8,7 @@ message Register {
   uint32 port_number = 1;
 }
 
+/// Request sent by the coordinator upon successful registration.
```
Nickster258, line 11: I have no idea how this got here. Thought I had handled all of these cases.

project3/participant/tasks/multicast_receiver_task.cpp
```diff
@@ -46,4 +44,10 @@ thread_pool::Task::Status MulticastReceiver::RunAtomic() {
 
 void MulticastReceiver::CleanUp() { close(messenger_fd_); }
 
+void MulticastReceiver::WriteLineToLog(const std::string& message) {
+  std::ofstream log_file;
```
djpetti: Might be worth just opening the log file once in `SetUp` and keeping the handle around.

project3/participant/tasks/multicast_receiver_task.h
```diff
@@ -31,16 +32,22 @@ class MulticastReceiver : public thread_pool::Task {
    * @param port the port this receiver will listen on
```
djpetti, line 32: Update docs

djpetti (approved):
Approved pending minor changes

PR #44, opened 2021-04-11
Enable timeouts for queue pops.

djpetti:
This should make it easier to cancel tasks that primarily
wait on a queue for input.

### DIFF
--- a/common/queue/CMakeLists.txt
+++ b/common/queue/CMakeLists.txt
@@ -1,1 +1,4 @@
 add_subdirectory(tests)
+
+add_library(queue INTERFACE)
+target_link_libraries(queue INTERFACE loguru)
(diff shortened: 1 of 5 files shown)


Nickster258 (approved):
Looks good! This will come in handy.

jake-chandler (approved):
Looks good. However, I'm currently not able to figure why we'd need a timeout enabled queue pop in this project. Do you have an idea about where this would be used?

djpetti:
> Looks good. However, I'm currently not able to figure why we'd need a timeout enabled queue pop in this project. Do you have an idea about where this would be used?

This would have come in handy in the last project, where we had tasks that spent all their time waiting for data on queues. If you want to cancel those tasks, you then have to send some dummy value over the queue to get it to wake up and realize that it should exit, which is annoying. With this functionality, the tasks can simply wait on the queue with a timeout, checking the cancellation status in between every wait. It's basically the same thing that we did with timeouts on sockets in project 2 to enable cancellation of tasks that read from sockets.

PR #45, opened 2021-04-16
Feature/p4 proto messages

jake-chandler:
The message encapsulation should now account for all messages a Bootstrap name server can receive. After reading through the project description again, I'm fairly certain that I've captured all messages we'll need.

### DIFF
--- a/project3/README.md
+++ b/project3/README.md
@@ -1,4 +1,4 @@
-# Building the Code
+m# Building the Code
 
 Project 3 code can be built on most Linux machines. See the repository README for general instructions on setting up
 the build environment.
--- a/project3/README.md
+++ b/project3/README.md
@@ -1,4 +1,4 @@
-m# Building the Code
+# Building the Code
 
 Project 3 code can be built on most Linux machines. See the repository README for general instructions on setting up
 the build environment.
--- a/project4/proto/CMakeLists.txt
+++ b/project4/proto/CMakeLists.txt
@@ -2,20 +2,20 @@ include(FindProtobuf)
 find_package(Protobuf REQUIRED)
 
 include_directories(${PROTOBUF_INCLUDE_DIR})
-PROTOBUF_GENERATE_CPP(PROTO_SRC PROTO_HEADER pub_sub_messages.proto)
-add_library(pub_sub_proto ${PROTO_HEADER} ${PROTO_SRC})
+PROTOBUF_GENERATE_CPP(PROTO_SRC PROTO_HEADER consistent_hash_msgs.proto)
+add_library(consistent_hash_msgs ${PROTO_HEADER} ${PROTO_SRC})
 # Instruct everything that links to this to also link to Protobuf.
-target_link_libraries(pub_sub_proto INTERFACE ${Protobuf_LIBRARIES})
+target_link_libraries(consistent_hash_msgs INTERFACE ${Protobuf_LIBRARIES})
 # Make sure everything that depends on this can find the headers.
-target_include_directories(pub_sub_proto INTERFACE ${CMAKE_CURRENT_BINARY_DIR})
+target_include_directories(consistent_hash_msgs INTERFACE ${CMAKE_CURRENT_BINARY_DIR})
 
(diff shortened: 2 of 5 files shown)


project3/README.md
```diff
@@ -1,4 +1,4 @@
-# Building the Code
+m# Building the Code
```
djpetti: I assume this was accidental?

project4/proto/CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
+include(FindProtobuf)
+find_package(Protobuf REQUIRED)
+
+include_directories(${PROTOBUF_INCLUDE_DIR})
+PROTOBUF_GENERATE_CPP(PROTO_SRC PROTO_HEADER pub_sub_messages.proto)
+add_library(pub_sub_proto ${PROTO_HEADER} ${PROTO_SRC})
```
djpetti: You will need a different name for this target. I'm actually surprised this builds as it is.

project4/proto/pub_sub_messages.proto
```diff
@@ -0,0 +1,140 @@
+syntax = "proto3";
+
+package pub_sub_messages;
```
djpetti: Call this something else.

project4/proto/pub_sub_messages.proto
```diff
@@ -0,0 +1,140 @@
...
+  /// the id of the entering name server
+  uint32 id = 1;
+}
+
+/// entrance information to be rippled through the name servers.
+message EntranceInformation {
+  /// the id of the entering name server
+  uint32 id = 1;
+  /// the id of the successor of the entering name server
+  uint32 successor_id = 2;
```
djpetti: (minor) Organizationally, it might be a little clearer if you encapsulated `id`, `ip`, and `port` in a sub-message.

project4/proto/pub_sub_messages.proto
```diff
@@ -0,0 +1,140 @@
...
+  /// the id of the predecessor's new successor.
+  uint32 id = 1;
+  /// the ip of the predecessor's new successor.
+  string ip = 2;
+  /// the port the predecessor's new successor is listening on.
+  uint32 port = 3;
+}
+/// encapsulates all messages that can be received by the bootstrap name server
+message BootstrapMessage {
+  oneof BootstrapMessages {
```
djpetti: (minor) I'd name this inner `oneof` something shorter, just because having a longer name will make the code more verbose. How about `messages`?

project4/proto/pub_sub_messages.proto
```diff
@@ -0,0 +1,140 @@
...
+message UpdateSuccessorRequest {
+  /// the id of the predecessor's new successor.
+  uint32 id = 1;
+  /// the ip of the predecessor's new successor.
+  string ip = 2;
+  /// the port the predecessor's new successor is listening on.
+  uint32 port = 3;
+}
+/// encapsulates all messages that can be received by the bootstrap name server
+message BootstrapMessage {
```
djpetti: Perhaps a tidier way to do this would be to have one of the options in the `oneof` be a `NameServerMessage`, and then everything else in the `oneof` are messages unique to the bootstrap node.

djpetti (changes requested):
I have no real qualms about the protocol itself. I suspect we'll discover any issues with that only when we start using it.

project4/proto/consistent_hash_msgs.proto
```diff
@@ -1,7 +1,16 @@
 syntax = "proto3";
 
-package pub_sub_messages;
+package consistent_hash_msgs;
 
+/// basic server information
+message ServerInfo {
+  /// the id of the name server's new predecessor
```
djpetti: You probably want these docs to be more general now. They should probably just refer to "the server" as opposed to predecessors and successors.

djpetti (approved):
One minor comment. Other than that, LGTM

project4/proto/consistent_hash_msgs.proto
jake-chandler: good catch; to be honest I just did a copy/paste and kinda forgot about it...

PR #46, opened 2021-04-17
Add client portion of message-passing library.

djpetti:
The client does exactly what it's name suggests: it connects to a server and allows for sending and receiving messages. However, it has the distinct advantage of eliminating all the socket and protobuf-related boilerplate.


common/message_passing/client.h
```diff
@@ -0,0 +1,154 @@
...
+ */
+class Client {
+ public:
+  /**
+   * @param thread_pool Thread pool to use internally for managing associated
+   *    tasks.
+   * @param destination The destination node that we will send messages to.
+   */
+  Client(std::shared_ptr<thread_pool::ThreadPool> thread_pool,
+                Endpoint destination);
```
jake-chandler: `Endpoint` seems very handy here

common/message_passing/client.h
```diff
@@ -0,0 +1,154 @@
...
+
+  /**
+   * @brief Receives a message.
+   * @tparam MessageType The type of message we expect.
+   * @param message[out] The message that we received.
+   * @return True if a message was successfully received and parsed, false
+   *    otherwise.
+   */
+  template <class MessageType>
+  bool Receive(MessageType* message) {
```
jake-chandler, line 64: Why does the implementation live in the header file? I'm guessing it's because of the templating...

common/message_passing/client.h
```diff
@@ -0,0 +1,154 @@
...
+  Endpoint endpoint_;
+  /**
+   * @brief Queue containing messages for the sender task to send.
+   */
+  std::shared_ptr<queue::Queue<SenderTask::SendQueueMessage>> send_queue_;
+  /// Queue containing received messages.
+  std::shared_ptr<queue::Queue<ReceiverTask::ReceiveQueueMessage>>
+      receive_queue_;
+  /// The task responsible for sending messages.
+  std::shared_ptr<SenderTask> sender_task_ = nullptr;
```
jake-chandler, line 140: Is there a reason you initialized them to `nullptr` rather than leaving them uninitialized?

common/message_passing/client.cpp
```diff
@@ -0,0 +1,182 @@
...
+    // Connect to the server.
+    LOG_S(INFO) << "Connecting to " << endpoint_.hostname << ":"
+                << endpoint_.port << "...";
+    const auto kAddress = MakeAddress(endpoint_.port);
+    client_fd_ = SetUpSocket(kAddress, endpoint_.hostname);
+
+    // Create the task for sending messages.
+    sender_task_ = std::make_shared<SenderTask>(
+        client_fd_, send_queue_, [this](MessageId id, int status) {
+          {
```
jake-chandler, line 165: Why are you encapsulating the two following lines of code in curly braces here?

jake-chandler (approved):
Looks great! I left a few comments just for some clarification.

common/message_passing/client.h
djpetti: Yep. Templated classes and functions, unfortunately, must be defined within the header file.

common/message_passing/client.h
djpetti: Sort of a defensive programming thing. If you initialize them to `nullptr` and try to do something with them, it's *guaranteed* to segfault as opposed to being potentially-undefined behavior. Obviously, this situation should never happen, and I'd rather catch it quickly if it ever does.

common/message_passing/client.cpp
djpetti: This is so the lock gets released before the condition variable `notify()` gets run. Technically, it wouldn't be a bug if it were run with the lock held, but since any thread waking up will need to immediately grab the lock anyway, it would degrade performance to do it that way.

Nickster258 (approved):

PR #47, opened 2021-04-24
Add server half of message passing code.

djpetti:
Again, the server class does exactly what it sounds like: it implements a server that listens on a port.

Interestingly, all the actual listening, as well as the handling of clients, is done in separate threads, so this should allow us to use a single-threaded implementation in situations where we were previously using a multi-threaded implementation.

This might not be entirely complete; I think there are still a few tests that I want to write. However, I think it's good enough to review/use.

### DIFF
--- a/common/message_passing/client.h
+++ b/common/message_passing/client.h
@@ -13,6 +13,7 @@
 #include <unordered_map>
 #include <vector>
 
+#include "node.h"
 #include "queue/queue.h"
 #include "tasks/receiver_task.h"
 #include "tasks/sender_task.h"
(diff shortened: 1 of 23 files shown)


common/message_passing/client.h
```diff
@@ -23,9 +24,9 @@
...
 
 /**
- * @brief Handles sending messages to another node.
+ * @brief Handles connecting to another node and sending/receiving messages.
  */
-class Client {
+class Client : public Node {
  public:
   /**
    * @param thread_pool Thread pool to use internally for managing associated
```
jake-chandler, line 32: I'm not sure what you mean by "managing associated"

jake-chandler (approved):
Looks straightforward to use. Thanks!

Nickster258 (approved):
Hard to critique this, looks really clean!

common/message_passing/client.h
djpetti: Each `Client` instance requires a `SenderTask` and a `ReceiverTask`, which are the "associated tasks" to be managed.

PR #48, opened 2021-04-24
Bootstrap, nameserver basic structure

Nickster258:
A few quick notes:
1. I feel a little guilty in reusing the ConsoleTask from P3. This should be moved / genericized, time permitting.
2. This was put together before fully realizing the message passing code, so some of the design ideas in this are subject to change.


project4/name_server/nameserver.h
```diff
@@ -0,0 +1,81 @@
...
+  /// Predecessor nameserver
+  std::map<std::string, int> predecessor;
+
+  /// Successor nameserver
+  std::map<std::string, int> successor;
+
+  /// Key bounds
+  std::pair<int, int> bounds;
+
+  /// Bootstrap server IP
```
jake-chandler: Maybe we could encapsulate the following variables into a similar structure as Daniel's `Endpoint`

project4/name_server/nameserver.h
```diff
@@ -0,0 +1,81 @@
...
+  void HandleRequest(const consistent_hash_msgs::UpdatePredecessorResponse &request);
+
+  void HandleRequest(const consistent_hash_msgs::LookUpResult &request);
+
+  void HandleRequest(const consistent_hash_msgs::InsertResult &request);
+
+  void HandleRequest(const consistent_hash_msgs::DeleteResult &request);
+
+  /**
+   * @param predecessor Predecessor if true, else successor
```
jake-chandler: Perhaps we could just have the predecessor/successor map object instead of the boolean?

project4/name_server/nameserver.h
```diff
@@ -0,0 +1,81 @@
...
+   * @param request The request to forward
+   */
+  void ForwardRequest(bool to_predecessor, const consistent_hash_msgs::NameServerMessage &request);
+
+ protected:
+  /// The key-value pairs in this nameserver
+  std::map<int, std::string> pairs;
+
+  /// Predecessor nameserver
+  std::map<std::string, int> predecessor;
```
jake-chandler: I think the types for the predecessors and successors should be a structure that encapsulates all the relevant info for connection, like the one described in the comment below.

What was the idea behind making the predecessor/successor a `map`?

project4/name_server/nameserver.h
```diff
@@ -0,0 +1,81 @@
...
+#include <map>
+#include <string>
+
+namespace nameserver {
+
+/**
+ * @class The basic name server
+ */
+class Nameserver {
+ public:
```
jake-chandler, line 24: Need to add functions for entering and exiting to be called by the driver

project4/name_server/nameserver.h
```diff
@@ -0,0 +1,81 @@
...
+
+  void HandleRequest(const consistent_hash_msgs::DeleteResult &request);
+
+  /**
+   * @param predecessor Predecessor if true, else successor
+   * @param request The request to forward
+   */
+  void ForwardRequest(bool to_predecessor, const consistent_hash_msgs::NameServerMessage &request);
+
+ protected:
```
jake-chandler: We will probably make helper functions that determine whether inserts, deletes, and inserts have been completed based off of message contents

jake-chandler (commented):
I think the overall structure will work. I've left some ideas in here that I want your thoughts on before we merge and split up tasks.

project4/name_server/nameserver.h
Nickster258: We could do that, but then it might as well be a static method. I was sorta thinking that this could be a "convenient" method to handle a oneshot requests to a predecessor or successor.

project4/name_server/nameserver.h
Nickster258: Oh that is just a silly typo. It is supposed to be a pair.

project4/name_server/nameserver.h
Nickster258: Yeah, we should probably use the `Endpoint` struct directly for this.

project4/name_server/nameserver.h
jake-chandler: I gotcha

jake-chandler (approved):

project4/name_server/bootstrap.h
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * @class Bootstrap server
+ */
+class Bootstrap : public Nameserver {
+ public:
+  /**
+   * @param the config file of this bootstrap
+   */
+  explicit Bootstrap(const std::string& config_file);
```
djpetti: (minor) Maybe use `std::filesystem::path` here?

project4/name_server/bootstrap.h
```diff
@@ -0,0 +1,33 @@
...
+
+  /**
+   * @param Generic BootstrapMessage request
+   */
+  void HandleRequest(const consistent_hash_msgs::BootstrapMessage &request);
```
djpetti: These functions could use an `@brief` section in their documentation.

project4/name_server/bootstrap_driver.h
```diff
@@ -0,0 +1,28 @@
...
+
+class BootstrapDriver {
+ public:
+  explicit BootstrapDriver(const std::string& config_file);
+
+  [[noreturn]] void Start();
+
+ private:
+  /// Pool and tasks
+  thread_pool::ThreadPool pool_;
```
djpetti: You're probably going to want to make this a `shared_ptr` as well, especially if you plan to use the `message_passing` library.

project4/name_server/commands.h
```diff
@@ -0,0 +1,25 @@
...
+ */
+#ifndef PROJECT4_COMMANDS_H
+#define PROJECT4_COMMANDS_H
+
+#include <map>
+#include <string>
+
+namespace nameserver {
+
+enum NameserverCommand { ENTER, EXIT };
```
djpetti: Use strongly-typed enums here, e.g. `enum class`.

project4/name_server/commands.h
```diff
@@ -0,0 +1,25 @@
...
+
+const std::map<std::string, NameserverCommand> nameserver_cmds = {
```
djpetti: I'd generally prefer `unordered_map` to normal `map`, because it has better performance in most cases. Also, that's what we use in most of the rest of the code.

project4/name_server/nameserver.h
```diff
@@ -0,0 +1,79 @@
...
+
+/**
+ * @class The basic name server
+ */
+class Nameserver {
+ public:
+  /**
+   * @param the config file of this name server
+   */
+  explicit Nameserver(const std::string& config_file);
```
djpetti: Same comment here about `std::filesystem::path`.

project4/name_server/nameserver.h
```diff
@@ -0,0 +1,79 @@
...
+
+  void HandleRequest(const consistent_hash_msgs::InsertResult &request);
+
+  void HandleRequest(const consistent_hash_msgs::DeleteResult &request);
+
+  /**
+   * @param predecessor Predecessor if true, else successor
+   * @param request The request to forward
+   */
+  void ForwardRequest(bool to_predecessor, const consistent_hash_msgs::NameServerMessage &request);
```
djpetti: Also could use an `@brief` section in the documentation.

project4/name_server/nameserver.h
```diff
@@ -0,0 +1,79 @@
...
+  void HandleRequest(const consistent_hash_msgs::UpdatePredecessorResponse &request);
+
+  void HandleRequest(const consistent_hash_msgs::LookUpResult &request);
+
+  void HandleRequest(const consistent_hash_msgs::InsertResult &request);
+
+  void HandleRequest(const consistent_hash_msgs::DeleteResult &request);
+
+  /**
+   * @param predecessor Predecessor if true, else successor
```
djpetti: For situations like these, it makes your code a lot more readable if you create an `enum` to represent predecessor vs. successor.

project4/name_server/nameserver.h
```diff
@@ -0,0 +1,79 @@
...
+   * @param request The request to forward
+   */
+  void ForwardRequest(bool to_predecessor, const consistent_hash_msgs::NameServerMessage &request);
+
+ protected:
```
djpetti: Are we sure all these things need to be protected? I would default to making them private, and move them to protected on a case-by-case basis, as the need arises.

project4/name_server/nameserver_driver.h
```diff
@@ -0,0 +1,32 @@
...
+  thread_pool::ThreadPool pool_;
+  std::shared_ptr<nameserver_tasks::ConsoleTask> console_task_;
+  std::shared_ptr<nameserver_tasks::NameserverTask> nameserver_task_;
+
+  /// Config file
+  const std::string config_file_;
+};
+
+}  // namespace nameserver
+#endif  // PROJECT4_NAMESERVER_DRIVER_H
```
djpetti: This file needs docstrings.

project4/name_server/bootstrap_driver.h
```diff
@@ -0,0 +1,28 @@
...
+
+#include "../../common/thread_pool/task.h"
+#include "../../common/thread_pool/thread_pool.h"
+
+#include "tasks/console_task.h"
+#include "tasks/bootstrap_task.h"
+
+namespace nameserver {
+
+class BootstrapDriver {
```
djpetti, line 17: Is it possible to unify this class with `NameserverDriver`? You could use DI to inject the actual `Nameserver` subclass that you want to manage. If not, I would recommend creating a common interface class that both `BootstrapDriver` and `NameserverDriver` inherit from.

project4/name_server/nameserver_main.cpp
```diff
@@ -0,0 +1,48 @@
...
+ */
+
+#include <loguru.hpp>
+
+#include <iostream>
+
+#include "nameserver_driver.h"
+#include "bootstrap_driver.h"
+
+void start_nameserver(char* config) {
```
djpetti, line 12: If you take my suggestion about unifying `BootstrapDriver` and `NameserverDriver`, some of this code can be simplified.

project4/name_server/tasks/CMakeLists.txt
```diff
@@ -0,0 +1,2 @@
+add_library(name_server_tasks_lib console_task.cpp bootstrap_task.cpp nameserver_task.cpp ../nameserver.cpp ../bootstrap.cpp)
```
djpetti: `nameserver.cpp` and `bootstrap.cpp` should not be included here.

project4/name_server/tasks/bootstrap_task.h
```diff
@@ -0,0 +1,32 @@
...
+
+#include <iostream>
+#include "thread_pool/task.h"
+#include "../bootstrap.h"
+#include "../commands.h"
+#include "../../../common/queue/queue.h"
+
+namespace nameserver_tasks {
+
+class BootstrapTask : public thread_pool::Task {
```
djpetti, line 19: Maybe add class docs?

project4/name_server/tasks/bootstrap_task.h
```diff
@@ -0,0 +1,32 @@
...
+#ifndef PROJECT4_BOOTSTRAP_TASK_H
+#define PROJECT4_BOOTSTRAP_TASK_H
+
+#include <iostream>
+#include "thread_pool/task.h"
+#include "../bootstrap.h"
+#include "../commands.h"
+#include "../../../common/queue/queue.h"
+
+namespace nameserver_tasks {
```
djpetti: I would recommend you either make this a sub-namespace (e.g. `nameserver::tasks`), or put it all in the `nameserver` namespace.

project4/name_server/tasks/console_task.cpp
```diff
@@ -0,0 +1,37 @@
...
+
+ConsoleTask::ConsoleTask(const std::string& prompt) : prompt_(prompt) {}
+
+thread_pool::Task::Status ConsoleTask::SetUp() {
+  std::cout << prompt_ << std::flush;
+  return thread_pool::Task::Status::RUNNING;
+}
+
+thread_pool::Task::Status ConsoleTask::RunAtomic() {
+  const auto first_message = console_message_queue_.Pop();
```
djpetti: This should be retooled using `PopTimed`. If the timeout expires, it should simply return `Status::RUNNING`. That will save us the necessity of doing the ugly hack that I had to do to get this task to exit correctly in project 3.

project4/name_server/tasks/console_task.h
```diff
@@ -0,0 +1,49 @@
...
+#ifndef PROJECT4_CONSOLE_TASK_H
+#define PROJECT4_CONSOLE_TASK_H
+
+#include <unistd.h>
+
+#include <cstdint>
+#include <iostream>
+#include <string>
+#include <vector>
+#include "../../../common/queue/queue.h"
```
djpetti: (minor) This can simply be included as `#include "queue/queue.h"`.

project4/name_server/tasks/console_task.h
```diff
@@ -0,0 +1,49 @@
...
+
+#include "thread_pool/task.h"
+
+namespace nameserver_tasks {
+
+/**
+ * @class The coordinator of all participant console output
```
djpetti: Update comment.

project4/name_server/tasks/nameserver_task.h
```diff
@@ -0,0 +1,32 @@
...
+
+#include <iostream>
+#include "thread_pool/task.h"
+#include "../nameserver.h"
+#include "../commands.h"
+#include "../../../common/queue/queue.h"
+
+namespace nameserver_tasks {
+
+class NameserverTask : public thread_pool::Task {
```
djpetti, line 19: Once again, can we unify the `NameserverTask` and `BootstrapTask` classes? See my previous comments about unifying the driver classes for details.

djpetti (changes requested):
Main comments here are related to mitigating code duplication.

project4/name_server/nameserver.h
Nickster258: I made them protected because I anticipated almost all of them being needed by the Bootstrap that inherits this.

project4/name_server/bootstrap_driver.h
Nickster258: I have to say, I'm not familiar with DI (dependency injection?). It may be easier to make a generic driver interface and have the `BootstrapDriver` and the `NamespaceDriver` inherit it.

project4/name_server/tasks/CMakeLists.txt
Nickster258: It complains without having it.

project4/name_server/bootstrap_driver.h
djpetti: Dependency injection, yes. It sounds fancy, but in this case, the idea is actually pretty simple. Mostly, it would amount to adding an argument to the constructor for `BootstrapDriver` that allowed you to pass a `Nameserver` subclass, which would then be used by the driver. I'm also a fan of this design pattern because it makes testing easier. My personal opinion is that in this case, it will also reduce code duplication more than writing an interface.

project4/name_server/tasks/CMakeLists.txt
djpetti: This circular dependency suggests that there is some larger design issue. It's hard to say without seeing the actual errors, but my guess is that this is caused by instantiating `Nameserver` and `Bootstrap` classes within the tasks. The easiest way to solve this problem would be to pass in a `shared_ptr` (or `unique_ptr`) of the `Nameserver` or `Bootstrap` instance that you want to use in the constructor of the relevant task. This also has the benefit of reducing coupling and making your code more testable.

project4/name_server/tasks/bootstrap_task.h
```diff
@@ -0,0 +1,40 @@
...
+  Status SetUp() override;
+
+  Status RunAtomic() override;
+
+ private:
+  /// Bootstrap
+  nameserver::Bootstrap bootstrap_;
+
+  /// Command queue
+  queue::Queue<std::pair<nameserver::BootstrapCommand, std::string>> cmd_queue_;
```
djpetti: This will almost certainly end up having to be a `shared_ptr`, because for a queue to be useful, it has to be shared between multiple places.

project4/name_server/bootstrap_driver.h
```diff
@@ -17,18 +16,20 @@ namespace nameserver {
...
+  explicit BootstrapDriver(std::filesystem::path config_file);
 
   /**
    * Start the user input loop
    */
   [[noreturn]] void Start();
 
  private:
+  /// Bootstrap
+  std::shared_ptr<nameserver::Bootstrap> bootstrap_;
```
djpetti, line 31: I'm not clear on why this is a `shared_ptr`, because it seems to me that it wouldn't really have to be shared with other classes. Are you planning to inject this in the constructor, or are you using it in some way that I'm not seeing?

project4/name_server/nameserver.cpp
```diff
@@ -5,14 +5,12 @@ Nameserver::Nameserver(const std::filesystem::path config_file) {
...
-bool Nameserver::Enter() {
-  return false;
-}
-void Nameserver::Exit() {
+bool Nameserver::Enter() { return false; }
+
+void Nameserver::Exit() {}
 
-}
 void Nameserver::HandleRequest(
```
djpetti: What's the point of a `HandleRequest` overload for a generic protobuf message? Since `NameServerMessage` is the container message, I'd think that this was right before.

project4/name_server/tasks/console_task.cpp
```diff
@@ -13,14 +13,19 @@ thread_pool::Task::Status ConsoleTask::SetUp() {
...
+  }
   // Clear prompt line in expectation of incoming console statement
   ClearLine();
 
   std::cout << first_message << std::endl;
   // Print any additional lines consecutively.
   while (!console_message_queue_.Empty()) {
-    std::cout << console_message_queue_.Pop() << std::endl;
+    std::string output;
+    console_message_queue_.PopTimed(kTimeout, &output);
```
djpetti: Nope, this one was fine. You can still use `Pop` here because you're checking `Empty()` before.

project4/name_server/tasks/console_task.h
```diff
@@ -40,9 +40,10 @@ class ConsoleTask : public thread_pool::Task {
    * clear the line of the console that the cursor is currently on
    */
   void ClearLine();
-
   queue::Queue<std::string> console_message_queue_{};
   std::string prompt_;
+
+  constexpr static const auto kTimeout = std::chrono::milliseconds(100);
```
djpetti, line 47: (minor) How about a docstring explaining what this timeout is for?

djpetti (changes requested):
I'm not sure if this is finished yet, but I'm reviewing it anyway.

project4/name_server/bootstrap_driver.h
Nickster258: The bootstrap is also being used by the `bootstrap_task`, so it will still have to be a shared pointer in this case.

project4/name_server/nameserver.cpp
Nickster258: The point of genericizing was to be able to override it in `Bootstrap` to handle different requests accordingly, because it can be a `NameServerMessage` for a nameserver, or a `BootstrapMessage` for a bootstrap. The `Bootstrap` would call `Nameserver::HandleRequest()` to handle anything else that isn't a `BootstrapMessage`.

PR #49, opened 2021-04-27
Add more tests for message-passing.

djpetti:


common/message_passing/tests/test_integration.cpp
```diff
@@ -0,0 +1,150 @@
...
+ * and produce a response from the server.
+ */
+TEST(MessagePassingIntegration, RequestResponse) {
+  // Arrange.
+  auto config = MakeConfig();
+
+  // Act.
+  // Send a request to the server.
+  const auto kTestRequest = MakeTestMessage();
+  ASSERT_TRUE(Retry([&]() { return config.client->Send(kTestRequest) > 0; }));
```
Nickster258, line 115: Honestly, this looks like magic. Where should I read up on this?

Nickster258 (approved):
LGTM

common/message_passing/tests/test_integration.cpp
djpetti: You mean [lambda expressions](https://en.cppreference.com/w/cpp/language/lambda)?

jake-chandler: Yeah, I'm not entirely sure what's going on here either

jake-chandler (approved):

common/message_passing/tests/test_integration.cpp
Nickster258: > You mean [lambda expressions](https://en.cppreference.com/w/cpp/language/lambda)?

Yeah I had assumed it was something like that, just didn't expect it to be called a lambda.

PR #50, opened 2021-04-28
Project 4 implementation

Nickster258:
There is still an issue with `Bootstrap::HandleRequest()` for `BootstrapMessage`s. There is a commented `FIXME` with further information.


project4/name_server/bootstrap.cpp
```diff
@@ -1,40 +1,200 @@
...
+    // the entering server will now be the last in the ring.
+    request.mutable_predecessor_info()->set_id(predecessor_id_);
+    request.mutable_predecessor_info()->set_ip(predecessor_.hostname);
+    request.mutable_predecessor_info()->set_port(predecessor_.port);
+  } else if (!request.successor_info().IsInitialized() && request.predecessor_info().IsInitialized()) {
+    // the entering server will now be the first in the ring.
+    request.mutable_successor_info()->set_port(successor_.port);
+    request.mutable_successor_info()->set_ip(successor_.hostname);
+    request.mutable_predecessor_info()->set_id(successor_id_);
+  } else if (request.mutable_predecessor_info()->IsInitialized() || request.mutable_successor_info()->IsInitialized()) {
```
jake-chandler: I think there might be a logical error here?

jake-chandler: There is. should be `!request.mutable_predecessor_info()->IsInitialized() && !request.mutable_successor_info()->IsInitialized()`

project4/name_server/bootstrap.cpp
```diff
@@ -1,40 +1,200 @@
...
-Bootstrap::Bootstrap(std::shared_ptr<thread_pool::ThreadPool> pool,
-                     std::shared_ptr<nameserver::tasks::ConsoleTask> console_task,
-                     const std::filesystem::path config_file)
-    : Nameserver(pool, console_task, config_file) {}
+Bootstrap::Bootstrap(
+    std::shared_ptr<thread_pool::ThreadPool> pool,
+    std::shared_ptr<nameserver::tasks::ConsoleTask> console_task, int port,
+    std::unordered_map<uint, std::string> kvs)
+    : Nameserver(pool, console_task, port,
+                 {"127.0.0.1", static_cast<uint16_t>(port)}) {
```
djpetti, line 11: Are we sure we want this address hard-coded?

project4/name_server/bootstrap.cpp
```diff
@@ -1,40 +1,200 @@
...
+  pairs_ = kvs;
```
djpetti: Any reason you don't use initialization lists here? For instance,
```c++
...
: Nameserver(pool, console_task, port,
                 {"127.0.0.1", static_cast<uint16_t>(port)}),
pairs_(kvs), bounds_({0, 1023}), successor_(bootstrap) {}
```

Some of these are even just constants and can probably be set in the .h file.

project4/name_server/bootstrap.cpp
```diff
@@ -1,40 +1,200 @@
...
-void Bootstrap::HandleRequest(const google::protobuf::Message& request) {
-  /// Handle as if BootstrapMessage, else call:
-  /// nameserver::Nameserver::HandleRequest(request);
+void Bootstrap::HandleRequest(
+    const consistent_hash_msgs::BootstrapMessage& request,
+    message_passing::Endpoint source) {
+  if (request.has_entrance_request()) {
+    HandleRequest(request.entrance_request(), source);
+  } else if (request.has_name_server_message()) {
+    /// FIXME if receiving a type of NameServerMessage, but listening for a
```
djpetti, line 25: Is this still an issue? I thought @jake-chandler said something about fixing it...

project4/name_server/bootstrap.cpp
```diff
@@ -1,40 +1,200 @@
...
+    client_ = std::make_unique<message_passing::Client>(threadpool_, successor_);
+    client_->Send(entrance_info);
+  }
 
-void Bootstrap::InitiateEntrance(const message_passing::Endpoint server) {}
+}
 
-std::string Bootstrap::LookUp(int key) { return ""; }
+void Bootstrap::HandleRequest(
+    consistent_hash_msgs::EntranceInformation& request) {
```
djpetti: Once again, I frown upon passing stuff by non-`const` reference. To be really safe, you might actually want to make a copy of the message, since the semantics of received messages are such that we usually don't expect them to change. On the other hand, you could make it a non-`const` pointer.

project4/name_server/bootstrap.cpp
```diff
@@ -1,40 +1,200 @@
...
+    request.mutable_predecessor_info()->set_port(predecessor_.port);
+  } else if (!request.successor_info().IsInitialized() && request.predecessor_info().IsInitialized()) {
+    // the entering server will now be the first in the ring.
+    request.mutable_successor_info()->set_port(successor_.port);
+    request.mutable_successor_info()->set_ip(successor_.hostname);
+    request.mutable_predecessor_info()->set_id(successor_id_);
+  } else if (request.mutable_predecessor_info()->IsInitialized() || request.mutable_successor_info()->IsInitialized()) {
+    // this should never happen.
+    // error
+    return;
```
djpetti, line 105: This should definitely be logged, probably at the `FATAL` if you truly never expect it to happen.

project4/name_server/bootstrap.cpp
```diff
@@ -1,40 +1,200 @@
...
+  }
+  client_ = std::make_unique<message_passing::Client>(threadpool_, entering_nameserver_);
+  if (client_->Send(request) < 0 ) {
+    // error
```
djpetti, line 111: This should also be logged. Furthermore, since you don't actually propagate this error to the caller anyway, I would recommend `SendAsync()`. This can still fail, but it will return faster, and if the actual send fails it will be logged by the message-passing library.

project4/name_server/bootstrap.cpp
```diff
@@ -1,40 +1,200 @@
...
+  }
+}
 
-void Bootstrap::Insert(int key, std::string val) {}
+void Bootstrap::LookUp(uint key) {
+  // key-val resides in this nameserver
+  auto itr = pairs_.find(key);
```
djpetti: (minor) Should this be `const`?

project4/name_server/bootstrap.cpp
```diff
@@ -1,40 +1,200 @@
...
+  } else if (successor_ == bootstrap_) {
+    console_task_->SendConsole("Key not found.");
+  } else {
+    consistent_hash_msgs::LookUpResult lookup;
+    lookup.set_id(0);
+    lookup.set_key(key);
+    client_ =
+        std::make_unique<message_passing::Client>(threadpool_, successor_);
+    if (client_->Send(lookup) < 0) {
+      // error
```
djpetti, line 131: See my previous comment about logging and `SendAsync`.

project4/name_server/bootstrap.cpp
```diff
@@ -1,40 +1,200 @@
...
+  } else if (successor_ == bootstrap_) {
+    console_task_->SendConsole("Key out of range.");
+  } else {
+    consistent_hash_msgs::InsertResult insert;
+    insert.set_id(0);
+    insert.set_key(0);
+    client_ =
+        std::make_unique<message_passing::Client>(threadpool_, successor_);
+    if (client_->Send(insert) < 0) {
+      // error
```
djpetti, line 154: See my previous comment about logging and `SendAsync`.

project4/name_server/bootstrap.cpp
```diff
@@ -1,40 +1,200 @@
...
+    }
+  }
+}
 
-void Bootstrap::HandleRequest(
-    const consistent_hash_msgs::EntranceInformation& request) {}
+void Bootstrap::Delete(uint key) {
+  // key-val resides in this nameserver
+  auto itr = pairs_.find(key);
```
djpetti, line 162: (minor) Should this be `const`?

project4/name_server/bootstrap.cpp
```diff
@@ -1,40 +1,200 @@
...
+  } else if (successor_ == bootstrap_) {
+    console_task_->SendConsole("Key not found.");
+  } else {
+    consistent_hash_msgs::DeleteResult delete_r;
+    delete_r.set_delete_success(false);
+    delete_r.set_key(key);
+    client_ =
+        std::make_unique<message_passing::Client>(threadpool_, successor_);
+    if (client_->Send(delete_r) < 0) {
+      // error
```
djpetti, line 176: See my previous comment about logging and `SendAsync`.

project4/name_server/bootstrap.cpp
```diff
@@ -1,40 +1,200 @@
...
+    }
+  }
+}
 
-void Bootstrap::HandleRequest(
-    const consistent_hash_msgs::ExitInformation& request) {}
+void Bootstrap::HandleRequest(consistent_hash_msgs::LookUpResult& request) {
```
djpetti: See my previous comment about non-`const` references.

project4/name_server/bootstrap.cpp
```diff
@@ -1,40 +1,200 @@
...
+  // denote this server as contacted
+  auto ids = request.server_ids();
+  request.set_server_ids(ids.size(), 0);
 
-void Bootstrap::HandleRequest(
-    const consistent_hash_msgs::LookUpResult& request) {}
+  if (request.value().empty()) {
+    // id 0 indicates bootstrap, except bootstrap never sends this request
```
djpetti: This comment seems a little confusing. Can you reword?

project4/name_server/bootstrap.cpp
```diff
@@ -1,40 +1,200 @@
...
-void Bootstrap::HandleRequest(
-    const consistent_hash_msgs::ExitInformation& request) {}
+void Bootstrap::HandleRequest(consistent_hash_msgs::LookUpResult& request) {
+  // denote this server as contacted
+  auto ids = request.server_ids();
+  request.set_server_ids(ids.size(), 0);
 
-void Bootstrap::HandleRequest(
-    const consistent_hash_msgs::LookUpResult& request) {}
+  if (request.value().empty()) {
```
djpetti: I thought `empty()` was only a valid method on repeated fields? I'm surprised this works...

project4/name_server/nameserver.cpp
```diff
@@ -1,41 +1,276 @@
...
+    HandleRequest(req);
+  } else if (msg.has_entrance_info()) {
+    auto req = msg.entrance_info();
+    HandleRequest(req);
+  } else if (msg.has_exit_info()) {
+    HandleRequest(msg.exit_info());
+  } else if (msg.has_update_pred_req()) {
+    auto req = msg.update_pred_req();
+    HandleRequest(req, source);
+  }
```
djpetti, line 61: If the message has none of these, that should probably be logged as an error.

project4/name_server/nameserver.cpp
```diff
@@ -1,41 +1,276 @@
...
+  consistent_hash_msgs::UpdatePredecessorRequest update_pred_req;
+  consistent_hash_msgs::UpdateSuccessorRequest update_succ_req;
+
+  // send entrance info request to bootstrap
+  if (!client_->SendRequest(entrance_req, &entrance_info)){ return false; }
+
+  if (entrance_info.IsInitialized()) {
+    // ring is not empty.
+    // update successor & predecessor information
+    successor_.port = entrance_info.mutable_successor_info()->port();
```
djpetti: (minor) This can be made more readable by simply using, e.g. `successor_info()` instead of `mutable_successor_info()`. The mutable version is not necessary because you're not changing it.

project4/name_server/nameserver.cpp
```diff
@@ -1,41 +1,276 @@
...
-bool Nameserver::Enter() { return false; }
+bool Nameserver::Enter() {
+  consistent_hash_msgs::EntranceRequest entrance_req;
+  consistent_hash_msgs::EntranceInformation entrance_info;
+  consistent_hash_msgs::UpdatePredecessorResponse update_pred_res;
+  consistent_hash_msgs::UpdatePredecessorRequest update_pred_req;
+  consistent_hash_msgs::UpdateSuccessorRequest update_succ_req;
+
+  // send entrance info request to bootstrap
+  if (!client_->SendRequest(entrance_req, &entrance_info)){ return false; }
```
djpetti: This is not safe, because `client_` seems to get reset every time you want to connect to a new one. I was going to comment about this somewhere else, but in general, I think there's not much point in making this a class member, since the predecessor and successor can change. Just use local `Client` instances when you want to send something. Then you can ditch the `unique_ptr`s too. The only place where it might be reasonable to have a `Client` as a class member is in the case of the `Client` connected to the bootstrap node, since that's fixed.

project4/name_server/nameserver.cpp
```diff
@@ -1,41 +1,276 @@
...
+  if (client_->Send(update_succ_req) < 0){ return false; }
 
-void Nameserver::Exit() {}
+  return true;
+}
+
+void Nameserver::Exit() {
+  consistent_hash_msgs::ExitInformation exit_info;
+  consistent_hash_msgs::UpdateSuccessorRequest update_succ_req;
+  int count = 0;
```
djpetti: (minor) Can this variable name be more descriptive?

project4/name_server/nameserver.cpp
```diff
@@ -1,41 +1,276 @@
...
+    count++;
+  }
+  exit_info.mutable_predecessor_info()->set_port(predecessor_.port);
+  exit_info.mutable_predecessor_info()->set_id(predecessor_id_);
+  exit_info.mutable_predecessor_info()->set_ip(predecessor_.hostname);
+  // send exit information to successor
+  // so it can update it's key's and predecessor info
+  client_ = std::make_unique<message_passing::Client>(threadpool_, successor_);
+  if (client_->Send(exit_info) < 0) {
+    // error
```
djpetti, line 158: See my previous comment about `SendAsync` and logging.

project4/name_server/nameserver.cpp
```diff
@@ -1,41 +1,276 @@
...
+  update_succ_req.mutable_successor_info()->set_id(successor_id_);
+  update_succ_req.mutable_successor_info()->set_port(successor_.port);
+  update_succ_req.mutable_successor_info()->set_ip(successor_.hostname);
 
-void Nameserver::HandleRequest(const google::protobuf::Message& request) {}
+  // tell predecessor to update it's successor info
+  client_ =
+      std::make_unique<message_passing::Client>(threadpool_, predecessor_);
+  if (client_->Send(exit_info) < 0) {
+    // error
```
djpetti, line 174: See my previous comment about `SendAsync` and logging.

project4/name_server/nameserver.cpp
```diff
@@ -1,41 +1,276 @@
...
+
+  // update predecessor info
+  predecessor_.port = request.predecessor_info().port();
+  predecessor_.hostname = request.predecessor_info().ip();
+  predecessor_id_ = request.predecessor_info().id();
+}
 
 void Nameserver::HandleRequest(
-    const consistent_hash_msgs::UpdatePredecessorRequest& request) {}
+    consistent_hash_msgs::UpdatePredecessorRequest& request,
```
djpetti: `request` can be `const`, right?

project4/name_server/nameserver.cpp
```diff
@@ -1,41 +1,276 @@
...
+  predecessor_.port = request.predecessor_info().port();
+  predecessor_.hostname = source.hostname;
+  predecessor_id_ = request.predecessor_info().id();
+
+  // give lower half of it's key-val range to the entering nameserver
+  consistent_hash_msgs::UpdatePredecessorResponse res;
+  res.set_lower_bounds(bounds_.first);
+  res.set_upper_bounds(bounds_.second / 2);
+  uint count = 0;
+  for (std::pair<uint, std::string> pair : pairs_) {
```
djpetti: (minor) It is a lot more convenient to just do `for (const auto& pair : pairs_) {...}`

project4/name_server/nameserver.cpp
```diff
@@ -1,41 +1,276 @@
...
+    res.set_keys(count, pair.first);
+    res.set_values(count, pair.second);
+    pairs_.erase(pair.first);
+    if (count == (bounds_.second / 2) - bounds_.first) {
```
djpetti: Is this the correct logic? Don't you want to be iterating through the whole map and checking whether `pair.first` is within the new bounds to decide whether to keep it or send it to the predecessor?

project4/name_server/nameserver.cpp
```diff
@@ -1,41 +1,276 @@
...
+      break;
+    }
+  }
+  bounds_.first = bounds_.second / 2 + 1;
```
djpetti: Is this supposed to be `bounds_.first += (bounds_.second - bounds_.first) / 2 + 1`?

project4/name_server/nameserver.cpp
```diff
@@ -1,41 +1,276 @@
...
+    message_passing::Endpoint source) {
+  // update predecessor info
+  predecessor_.port = request.predecessor_info().port();
+  predecessor_.hostname = source.hostname;
+  predecessor_id_ = request.predecessor_info().id();
+
+  // give lower half of it's key-val range to the entering nameserver
+  consistent_hash_msgs::UpdatePredecessorResponse res;
+  res.set_lower_bounds(bounds_.first);
+  res.set_upper_bounds(bounds_.second / 2);
```
djpetti: Is this supposed to be `res.set_upper_bounds(bounds_.first + (bounds_.second - bounds.first) / 2)` It might be more readable to do something like `const uint kMiddleKey = bounds_.first + (bounds_.second - bounds_.first) / 2;` and then use variable.

project4/name_server/nameserver.cpp
```diff
@@ -1,41 +1,276 @@
...
+    if (count == (bounds_.second / 2) - bounds_.first) {
+      break;
+    }
+  }
+  bounds_.first = bounds_.second / 2 + 1;
+
+  client_ =
+      std::make_unique<message_passing::Client>(threadpool_, predecessor_);
+  if (client_->Send(res) < 0) {
+    // error
```
djpetti, line 285: See my previous comment about logging and `SendAsync`.

project4/name_server/nameserver.cpp
```diff
@@ -1,41 +1,276 @@
...
+  }
+}
 
 void Nameserver::HandleRequest(
-    const consistent_hash_msgs::UpdateSuccessorRequest& request) {}
+    const consistent_hash_msgs::UpdateSuccessorRequest& request) {
+  // update successor info
```
djpetti, line 292: As a general rule, it would be nice to see more logging in the request handlers. At least maybe log that a particular type of message has been received... That will make integration a lot easier.

project4/name_server/nameserver.cpp
```diff
@@ -1,41 +1,276 @@
...
+  successor_.hostname = request.successor_info().ip();
+  successor_.port = request.successor_info().port();
+  successor_id_ = request.successor_info().id();
+}
 
 void Nameserver::HandleRequest(
```
djpetti, line 298: This is safe to ignore?

project4/name_server/nameserver.cpp
```diff
@@ -1,41 +1,276 @@
...
     const consistent_hash_msgs::UpdatePredecessorResponse& request) {}
 
-void Nameserver::HandleRequest(
-    const consistent_hash_msgs::LookUpResult& request) {}
+void Nameserver::HandleRequest(consistent_hash_msgs::LookUpResult& request) {
```
djpetti: See my previous comment about non-`const` references.

djpetti: See my previous comment about non-`const` references.

project4/name_server/nameserver.cpp
```diff
@@ -1,41 +1,276 @@
...
+  // denote this server as contacted
+  auto ids = request.server_ids();
+  request.set_server_ids(ids.size(), id_);
 
-void Nameserver::HandleRequest(
-    const consistent_hash_msgs::InsertResult& request) {}
+  // forward LookUpResult to successor
+  client_ = std::make_unique<message_passing::Client>(threadpool_, successor_);
+  if (client_->Send(request) < 0) {
+    // error
```
djpetti, line 319: See my previous comment about logging and `SendAsync`.

project4/name_server/nameserver.cpp
```diff
@@ -1,41 +1,276 @@
...
+    auto itr = pairs_.find(request.key());
+    if (itr != pairs_.end()) {
+      // value found
+      request.set_value(itr->second);
+      request.set_id(id_);
+    }
+  }
+  // denote this server as contacted
+  auto ids = request.server_ids();
+  request.set_server_ids(ids.size(), id_);
```
djpetti: I think the method that you want here is `add_server_ids()`. `set_server_ids()` might actually be undefined behavior since your index is technically out-of-bounds.

project4/name_server/nameserver.cpp
```diff
@@ -1,41 +1,276 @@
...
+void Nameserver::HandleRequest(consistent_hash_msgs::InsertResult& request) {
+  if (request.key() <= bounds_.second && request.key() >= bounds_.first) {
+    // key-val should be inserted in this nameserver
+    std::pair<int, std::string> pair(request.key(), request.value());
+    pairs_.insert(pair);
+    request.set_id(id_);
+  }
+  // denote this server as contacted
+  auto ids = request.server_ids();
+  request.set_server_ids(ids.size(), id_);
```
djpetti: See my previous comment about `add_server_ids()`.

project4/name_server/nameserver.cpp
```diff
@@ -1,41 +1,276 @@
...
+  // forward LookUpResult to successor
+  client_ = std::make_unique<message_passing::Client>(threadpool_, successor_);
+  if (client_->Send(request) < 0) {
+    // error
+  }
+}
 
-void Nameserver::HandleRequest(
-    const consistent_hash_msgs::DeleteResult& request) {}
+void Nameserver::HandleRequest(consistent_hash_msgs::InsertResult& request) {
```
djpetti: See my previous comment about non-`const` references.

project4/name_server/nameserver.cpp
```diff
@@ -1,41 +1,276 @@
...
+    request.set_id(id_);
+  }
+  // denote this server as contacted
+  auto ids = request.server_ids();
+  request.set_server_ids(ids.size(), id_);
+
+  // forward LookUpResult to successor
+  client_ = std::make_unique<message_passing::Client>(threadpool_, successor_);
+  if (client_->Send(request) < 0) {
+    // error
```
djpetti, line 343: See my previous comment about logging and `SendAsync`.

project4/name_server/nameserver.cpp
```diff
@@ -1,41 +1,276 @@
...
+  }
+}
+
+void Nameserver::HandleRequest(consistent_hash_msgs::DeleteResult& request) {
```
djpetti: See my previous comment about non-`const` requests.

project4/name_server/nameserver.cpp
```diff
@@ -1,41 +1,276 @@
...
+    auto itr = pairs_.find(request.key());
+    if (itr != pairs_.end()) {
+      // key-val pair found , now delete
+      pairs_.erase(itr);
+      request.set_delete_success(true);
+    }
+  }
+  // denote this server as contacted
+  auto ids = request.server_ids();
+  request.set_server_ids(ids.size(), id_);
```
djpetti: See my previous comment about `add_server_ids()`.

project4/name_server/nameserver.cpp
```diff
@@ -1,41 +1,276 @@
...
+
+  // forward LookUpResult to successor
+  client_ = std::make_unique<message_passing::Client>(threadpool_, successor_);
+  if (client_->Send(request) < 0) {
+    // error
```
djpetti, line 370: See my previous comment about logging and `SendAsync`.

project4/name_server/nameserver.h
```diff
@@ -28,7 +28,8 @@ class Nameserver {
    */
   Nameserver(std::shared_ptr<thread_pool::ThreadPool> pool,
```
djpetti, line 32: Update docs.

project4/name_server/tasks/nameserver_task.cpp
```diff
@@ -4,18 +4,16 @@
...
 
-/// TODO Set up socket listening
-thread_pool::Task::Status NameserverTask::SetUp() {
-  return Task::SetUp();
-}
-
-/// TODO Listen to all requests. Pass them accordingly to HandleResponse
-/// If there is a command in cmd_queue_, parse and execute.
 thread_pool::Task::Status NameserverTask::RunAtomic() {
+  message_passing::Endpoint endpoint;
```
djpetti: We should really combine these into one class. Have you tried simply substituting a `NameserverTask` where you would otherwise use a `BootstrapTask`? I bet this would work, since `Bootstrap` is a subclass of `Nameserver`.

djpetti (changes requested):
I know this looks like a scary number of comments, but many of them are the same comment pointed out at multiple locations.

I see some things that look like bugs, some of which could be quite serious, so I'm requesting changes.

project4/name_server/tasks/nameserver_task.cpp
```diff
@@ -4,18 +4,16 @@
...
+  consistent_hash_msgs::NameServerMessage ns_msg;
+  if (nameserver_->server_->Receive(kTimeout, &ns_msg, &endpoint)) {
```
djpetti: From a design perspective, having the `NameserverTask` directly access the `server_` member is pretty tight coupling and would generally be frowned upon. Why not add a `ReceiveAndHandleRequest()` method to the `Nameserver` class that does exactly that? Otherwise, the `Server` should be a member of the `NameserverTask` class, IMO.

project4/name_server/bootstrap.cpp
jake-chandler: Yeah, I'm not exactly sure. I do know it popped up on the CLion autocomplete so I figured I'd give it a try. It would make sense for it to have an `empty()`, kinda like how messages have the `IsInitialized`

project4/name_server/nameserver.cpp
jake-chandler: ah, forgot about `pairs_` being an `unordered_map`

project4/name_server/nameserver.cpp
jake-chandler: ah yes, thanks for catching that. math is hard.

project4/name_server/nameserver.cpp
jake-chandler: yup. thanks again.

project4/name_server/nameserver.cpp
jake-chandler: I'm not sure what you're referring to.

project4/name_server/tasks/nameserver_task.cpp
Nickster258: I like the sound of that idea.

project4/name_server/tasks/nameserver_task.cpp
Nickster258: You're right, combining these would be best. I haven't yet tried it that way, but it is likely doable.

project4/name_server/bootstrap.cpp
Nickster258: Yeah, I just kept adding onto the constructor and never reevaluated it. Initialization lists and making these constant are the best way to go.

project4/name_server/bootstrap.cpp
Nickster258: I don't necessarily see a problem in doing it this way as the bootstrap, when starting, will only be aware of itself.

project4/name_server/bootstrap.cpp
Nickster258: From what I understand, yes.

project4/name_server/bootstrap.cpp
djpetti: Ah, the reason this works is because `value` is an `std::string`, so this is not going to do exactly what you want it to. However, if you assume that no user will ever set a value to an empty string, then it will work, but I'm not sure we can assume that. You might need to add a separate `is_found` value to the message.

project4/name_server/nameserver.cpp
djpetti: This function does nothing, so this message type is effectively ignored.

project4/name_server/bootstrap.cpp
djpetti: Yeah, I guess so

project4/name_server/nameserver.cpp
jake-chandler: Oh, so, the only time a nameserver would receive an `UpdatePredecessorResponse` would be during the entering process, when `Enter()` is being called. You'll notice in `Enter()` that the entering nameserver sends an `UpdatePredecessorRequests` and expects a response. That being said, I think this function can be deleted.

project4/name_server/bootstrap.cpp
jake-chandler: I'm confused about what the issue is, honestly, I'm not fully following the comment

project4/name_server/bootstrap.cpp
```diff
@@ -1,40 +1,218 @@
...
+    std::pair<int, std::string> pair(key, val);
+    pairs_.insert(pair);
+    console_task_->SendConsole("Inserted into: 0");
+    console_task_->SendConsole("Contacted: 0");
+  } else if (successor_ == bootstrap_) {
+    console_task_->SendConsole("Key out of range.");
+  } else {
+    consistent_hash_msgs::InsertResult insert;
+    insert.set_id(0);
+    insert.set_key(0);
```
jake-chandler: shouldn't this be `set_key(key)` ?

project4/name_server/bootstrap.cpp
```diff
@@ -1,40 +1,218 @@
...
+    // key-val should be inserted in this bootstrap
+    std::pair<int, std::string> pair(key, val);
+    pairs_.insert(pair);
+    console_task_->SendConsole("Inserted into: 0");
+    console_task_->SendConsole("Contacted: 0");
+  } else if (successor_ == bootstrap_) {
+    console_task_->SendConsole("Key out of range.");
+  } else {
+    consistent_hash_msgs::InsertResult insert;
+    insert.set_id(0);
```
jake-chandler, line 148: I think here you're supposed to be calling `set_value(val)` instead. For `InsertResults`, the `id` is only set by the nameserver that the key is inserted into.

project4/name_server/bootstrap.cpp
Nickster258: Oh, yeah it definitely should be.

project4/name_server/bootstrap.cpp
Nickster258: Actually I can't get that initialization list to work, or any other variant I try.

project4/name_server/bootstrap.cpp
Nickster258: I added `set_value()` but I kept `set_id()` because, when the bootstrap sees the message later, if `id` is `0`, the bootstrap knows it was not inserted into any nameserver.

project4/name_server/bootstrap.cpp
```diff
@@ -49,43 +52,59 @@ void Bootstrap::HandleRequest(
...
     // send empty entrance info message.
-    // the joining nameserver will know it's predecessor and successor are the bootstrap server.
+    // the joining nameserver will know it's predecessor and successor are the
+    // bootstrap server.
     client_->Send(entrance_info);
   } else {
     // send entrance info message to be filled out
     entrance_info.set_id(request.id());
-    client_ = std::make_unique<message_passing::Client>(threadpool_, successor_);
+    client_ =
```
djpetti: Is there a reason why `client_` is a class member? I still think this is better off as a local variable.

project4/name_server/bootstrap.cpp
```diff
@@ -49,43 +52,59 @@ void Bootstrap::HandleRequest(
   }
 }
 
+void Bootstrap::ReceiveAndHandle() {
```
djpetti, line 55: AFAIK, this is identical to the version from `Nameserver`, so I think you can just delete this.

project4/name_server/bootstrap.h
```diff
@@ -35,7 +35,7 @@ class Bootstrap : public Nameserver {
    * @param key The integer key for the corresponding value
    * @param val The value
    */
-  void Insert(uint key, std::string val);
+  void Insert(uint key, const std::string &val);
```
djpetti, line 40: `Nameserver` needs virtual destructor, and this class should override it. That might have been what `clang` was complaining about for the Mac build. Adding `~Bootstrap() override = default;` should do the trick.

project4/name_server/nameserver.h
```diff
@@ -46,20 +45,19 @@ class Nameserver {
...
 
-  /**
-   * @brief Handles a NameserverMessage
-   * @param msg the message
-   */
-  void HandleRequest(const consistent_hash_msgs::NameServerMessage &msg, const message_passing::Endpoint source);
-
-  /// The server object
-  std::unique_ptr<message_passing::Server> server_;
+  virtual /**
```
djpetti: (minor) This is some really weird formatting.

project4/name_server/nameserver.h
```diff
@@ -46,20 +45,19 @@ class Nameserver {
    */
   void Exit();
```
djpetti, line 48: This class should also have a virtual destructor. Again, something like `virtual ~Nameserver() = default` should be good.

djpetti (approved):
Approving pending resolution of comments.

project4/name_server/bootstrap.cpp
Nickster258: It listens to a different message type, so `Bootstrap` has to override it here.

project4/name_server/nameserver.h
Nickster258: This was unintentional. Fixed.

project4/name_server/bootstrap.cpp
djpetti: Ah, I see. If you wanted to get really fancy, you could make a templated protected version of this method and then just have the public method call it with the appropriate type parameter, but I'm not going to request that change.

